PR #122, opened 2015-07-09
Resets Default Block Event for Additional Flowcharts

bjennings76:
Added a hidden serialized `version` field, an `UpdateVersion()` function, and an overridable `Initialize()` function to the Flowchart component that allow for custom upgrades from previous versions as well as a one-time initialization after each flowchart instance creation.

Updated the default initialization function to reset the event handler of the default block on the newly created flowchart if previous flowcharts are found in `cachedFlowcharts` and the default block uses the default block name (to avoid resetting older previously existing flowcharts).

image

### DIFF
--- a/Assets/Fungus/Flowchart/Scripts/Flowchart.cs
+++ b/Assets/Fungus/Flowchart/Scripts/Flowchart.cs
@@ -20,6 +20,11 @@ namespace Fungus
 		 */
         public const string CURRENT_VERSION = "1.0";
 
+        /**
+        * The name of the initial block in a new flowchart.
+        */
+        public const string DEFAULT_BLOCK_NAME = "New Block";
+
 		/**
 		 * Variable to track flowchart's version and if initial set up has completed.
 		 */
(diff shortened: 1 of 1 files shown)


chrisgregan:
Thanks for the pull request. Looks like a great addition, I'll merge it during the week once I get a chance to test it out.

bjennings76:
Sure thing. I just noticed the problem called out in the introduction video and thought I could help. Be sure to test it out with existing projects. That's one thing I didn't have available when testing it out.

I also have a simpler solution that only works with the New Flowchart menu command (which means it wouldn't work if you dragged in a Flowchart prefab) if this one looks too problematic. This solution is more universal and also has some options for upgrading to future features.

chrisgregan:
Tested it out here and it works great. I agree this is the right way to do it rather than modifying the New Flowchart menu code as it works in both spawning cases. I can definitely see the UpdateVersion method coming in handy to maintain backwards compatibility.

Thanks for a nice workflow improvement!

bjennings76:
Sure thing! I'm glad I could help. Thanks for creating and sharing such a
great tool!

PR #124, opened 2015-07-26
Hovering over a click-enabled Clickable2D GameObject can change the c…

Zeouterlimits:
Hovering over a click-enabled Clickable2D GameObject can change the cursor to indicate the object is clickable.

By default hoverOverCursorTexture will be null, so there will be no visible change to existing behaviour.

Ye probably have better ideas for the Start() call, am open to suggestions.
Will create post in the forum too.


chrisgregan:
I made a few additions and integrated in another pull request. I extended the cursor changing to also work with draggable objects and added a Set Mouse Cursor command to set the default mouse cursor. Thanks again for the contribution!

PR #324, opened 2015-12-11
Set say dialog for character

FeniXb3:
The idea of this feature is simple - allow setting default say dialog for Character. It's useful when you have a long dialog with multiple characters with dedicated say dialogs - it's more comfortable to set only character for each say command and not set say dialog for each one explicite in the command.


chrisgregan:
Sorry it took so long to merge this in. Nice improvement!

Issue #231, opened 2015-09-21
Add line numbers in block command list


RonanPearce:
Add a property to the Flowchart object to show line numbers in the command list. This can easily be done by prepending the command index to the command summary string before displaying it.

chrisgregan:
Done.

PR #412, opened 2016-03-25
Add line numbers in block command list #231 and Use Unity Object picker for Say command

hawmalt:


chrisgregan:
Hi hawmalt, thanks for the pull request!

Just in general, it's best to send in each change as its own pull request so it's easier for me to comment on each change and merge them individually. When they're all mixed together it's harder to figure out which changes go with which feature.

Another thing to bear in mind is backwards compatibility, not just with earlier versions of Fungus but also with tutorial videos, documentation and articles that are out in the wild. This means we need to be careful not to change the visual interface too much, unless it's for a very good reason.

So here's my feedback on the changes.

1. Command List Line Numbers. It's a nice improvement but there are a couple issues. Looks like it doesn't handle the case where there are more than 99 commands in a list (assumes a number width of 20px). Might be better to prepend the index to the command summary instead? I'd also make line numbers optional by adding a 'Line Numbers' property to the Flowchart component which defaults to off. That way if you want them you can switch them on.

2. Using the GetVariable() methods is more robust, nice improvement! I would still leave the GetIntegerVariable(), etc. accessor methods in place though to help maintain backwards compatibility. I try not to remove functions or change signatures unless there's a compelling reason to do it.

3. Using an object picker instead of a dropdown menu for the Say command. Again this is going to cause issues for people following tutorial videos, etc. Many people using Fungus have never made a game or coded before so they can get lost easily. Unless there's a major benefit to using an object picker over the dropdown I would leave it as it is.

4. InvokeMethod. I'm not sure what's changed here, can you give me a description of the improvement?

Thanks again for taking the time to contribute your code, with a few tweaks I'd be happy to accept some of the changes.

hawmalt:
1. I prefixed the command summary with the command index.  Did not add a 'Line Numbers' property.

2. I re-implemented the accessor functions

3. I re-added the drop down menu but kept the object picker so that you can still use Unity's search bar function when finding a character.

4. Invoke method was changed to use GetVariable in GetParameterValues and SetVariable functions instead of the GetBooleanVariable etc.

PR #459, opened 2016-04-21
Add "Play from Selected Command" to the BlockEditor ContextMenu

inarizushi:
In the Block Editor, when you have only one command selected and you're in play mode, it will give you the option to play from the selected command. If it's from a block that isn't executing, it will stop the current executing block then start the selected one.

### DIFF
--- a/Assets/Fungus/Flowchart/Editor/BlockEditor.cs
+++ b/Assets/Fungus/Flowchart/Editor/BlockEditor.cs
@@ -1027,20 +1027,16 @@ namespace Fungus
             Block targetBlock = target as Block;
             Flowchart flowchart = targetBlock.GetFlowchart();
             Command command = flowchart.selectedCommands[0];
-            targetBlock.jumpToCommandIndex = command.commandIndex;
             if (!targetBlock.IsExecuting())
             {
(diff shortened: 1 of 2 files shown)


inarizushi:
I had been having some trouble with the latest from master. I keep getting the error: Assets/Fungus/Thirdparty/FungusLua/Scripts/LuaUtils.cs(403,17): error CS0103: The name `timeScale' does not exist in the current context
when I change it to Time.timescale the errors go away. 

Is it a problem with the version of unity I'm running: 5.3.3?

chrisgregan:
Whoops - forgot to commit a change. I've updated master.

inarizushi:
Sorry, I have two github accounts. The other one is me too. 

So the latest commit works. There's still more work I want to do with it though. I couldn't get targetBlock.Execute(command.commandIndex); to work. When I looked in the debugger, it was like it never reached the Execute command. That's also why I had to manually set jumpToCommandIndex instead of passing it into Execute. Can I blame this all on coroutines?

chrisgregan:
Oh yeah, you need to start a coroutine if you want to use the Execute method directly. You can use the Block object to do that, like this. Best to use the proper method I think in this case.
```
Block b; 
b.StartCoroutine(b.Execute(10));
```

PR #461, opened 2016-04-23
Stop all executing blocks on play selected

inarizushi:
Quick bugfix: If there is another block executing, and then you use the "play selected" command on a non-executing block, both blocks will execute at the same time. I changed it so that it will stop all executing blocks before executing the selected block.


chrisgregan:
Hmm.. I can see why this is a good default, but you might want parallel executing Blocks though sometimes? Fungus is designed so that you can have multiple Blocks executing in a Flowchart at the same time, it's handy for lots of things.

What about if there were two menu options? "Play from selected" and "Stop all and play"? That way you can choose which you want.

inarizushi:
Woops didn't know that was the case. I like the idea of the two different options. I'll get to work on that. Thanks!

inarizushi:
So like this?

chrisgregan:
Perfect - thanks!

PR #476, opened 2016-05-16
No Music clip selected error

kalenobrien15:
Changed to 'No sound clip selected', could make both more universal and say no audio clip selected.


chrisgregan:
Cheers Kal!

PR #487, opened 2016-05-30
Fade ui bug

chrisgregan:


chrisgregan:
Created this pr by accident - deleting it

PR #490, opened 2016-06-09
Feature/portrait system refactor

lealeelu:
Move most of the logic of the Portrait command into a Portrait Controller.

### DIFF
--- a/Assets/Fungus/Narrative/Resources/Stage.prefab
+++ b/Assets/Fungus/Narrative/Resources/Stage.prefab
@@ -43,6 +43,7 @@ GameObject:
   m_Component:
   - 4: {fileID: 410270}
   - 114: {fileID: 11410274}
+  - 114: {fileID: 11453434}
   m_Layer: 0
   m_Name: Stage
   m_TagString: Untagged
@@ -331,6 +332,17 @@ MonoBehaviour:
   m_FillAmount: 1
   m_FillClockwise: 1
   m_FillOrigin: 0
+--- !u!114 &11453434
+MonoBehaviour:
+  m_ObjectHideFlags: 1
+  m_PrefabParentObject: {fileID: 0}
+  m_PrefabInternal: {fileID: 100100000}
+  m_GameObject: {fileID: 110274}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: cc7c4e379c0d3bb4eb7537ad653ea7f6, type: 3}
+  m_Name: 
+  m_EditorClassIdentifier: 
 --- !u!222 &22210270
 CanvasRenderer:
   m_ObjectHideFlags: 1
--- a/Assets/Fungus/Narrative/Resources/Stage.prefab
+++ b/Assets/Fungus/Narrative/Resources/Stage.prefab
@@ -43,7 +43,6 @@ GameObject:
   m_Component:
   - 4: {fileID: 410270}
   - 114: {fileID: 11410274}
-  - 114: {fileID: 11453434}
   m_Layer: 0
   m_Name: Stage
   m_TagString: Untagged
@@ -332,17 +331,6 @@ MonoBehaviour:
   m_FillAmount: 1
   m_FillClockwise: 1
   m_FillOrigin: 0
---- !u!114 &11453434
-MonoBehaviour:
-  m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
-  m_PrefabInternal: {fileID: 100100000}
-  m_GameObject: {fileID: 110274}
-  m_Enabled: 1
-  m_EditorHideFlags: 0
-  m_Script: {fileID: 11500000, guid: cc7c4e379c0d3bb4eb7537ad653ea7f6, type: 3}
-  m_Name: 
-  m_EditorClassIdentifier: 
 --- !u!222 &22210270
 CanvasRenderer:
   m_ObjectHideFlags: 1
@@ -388,7 +376,6 @@ Canvas:
   m_ReceivesEvents: 1
   m_OverrideSorting: 0
   m_OverridePixelPerfect: 0
-  m_SortingBucketNormalizedSize: 0
   m_SortingLayerID: 0
   m_SortingOrder: 0
   m_TargetDisplay: 0
--- a/Assets/Fungus/Narrative/Scripts/Character.cs
+++ b/Assets/Fungus/Narrative/Scripts/Character.cs
@@ -54,6 +54,12 @@ namespace Fungus
 			nameText = standardText;
 		}
 
+		/// <summary>
+		/// Looks for a portrait by name on a character
+		/// If none is found, give a warning and return a blank sprite
+		/// </summary>
+		/// <param name="portrait_string"></param>
+		/// <returns>Character portrait sprite</returns>
         public virtual Sprite GetPortrait(string portrait_string)
         {
             if (portrait_string == null)
--- a/Assets/Fungus/Narrative/Scripts/Character.cs
+++ b/Assets/Fungus/Narrative/Scripts/Character.cs
@@ -56,6 +56,13 @@ namespace Fungus
 
         public virtual Sprite GetPortrait(string portrait_string)
         {
+            if (portrait_string == null)
+            {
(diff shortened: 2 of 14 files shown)


lealeelu:
Not done yet! 
I'm having trouble with the Continue() from Portrait.cs. I think it's messing up the Narrative/PortraitTest.

chrisgregan:
Looks good, going in the right direction for sure. Continue() needs be called at the correct times back on the Portrait command. Easiest way to do this would be via a delegate, Action would do the trick - like this.

```
stage.getPortraitController().RunPortraitCommand(options, () => {
     Continue();
});

public void RunPortraitCommand(PortraitOptions options, Action onComplete)
{
...
    onComplete();
    return;
...
}
```

lealeelu:
Cool, I'll add that tonight.

lealeelu:
The tests pass now but they look a little strange. Sometimes the portrait turn entirely white.

lealeelu:
Everything seems to be back in working order. If it looks good to you, I'm going to start working on helper functions.

lealeelu:
What's a good way to test the helper functions? Is there anything else I need to do before I move onto lua and documentation?

chrisgregan:
Looks great! I'll do a proper line by line review over the weekend and commit, but on first pass through look it looks spot on.

You can add some new tests yourself for the new helper functions, add a new scene under the Tests/Narrative folder. The Create button on the Integration Test window will create a new test. I find it easiest these days to use Lua to perform the tests, but you can also use the IntegrationTest component or a c# script - whatever you prefer.

I usually use the assets from the Sherlock demo to test portrait system, but you can add your own if you like (assuming they're properly open source licensed).

lealeelu:
Yay! Thanks!
So this is what I think I have left to do:
- [ ] Figure out what to do with the FinishPortraitCommand bit - it's buggy
- [ ] Finish Lua Stage Control example.
- [ ] Create new Scene in Tests/Narrative folder with Lua tests of the helper functions
- [ ] Write some documentation on how to use it and point to the example project - but this needs to be added to the documentation repository

I'm so glad to see it mostly working ^_^

chrisgregan:
Sounds good. If you want to update the docs send in a pull request on the [fungus-site](https://github.com/snozbot/fungus-site) repo and then I'll pull an update to the website. Thanks for all the hard work!

lealeelu:
Hope I didn't mess up anything with that push. I did a pull rebase with the master branch to catch up.

lealeelu:
I'm having a heck of a time figuring out why the hide command is so spastic. I put in an example at FungusExamples/FungusLua/Narrative if you want to see what I mean. 

I think I'll punt on this for a day or so if you want to take a crack at it.

lealeelu:
nvm figured it out. Struct variables automatically get assigned values that I didn't expect:
null for reference types, 0 for numeric types, false for bools, etc.

chrisgregan:
Cool, prob be Monday when I get to look at merging this in

lealeelu:
Ok, for real this time I'm going to put this down and work on something else for the next couple days. Give you some time to make heads or tails of it.

Here's what else I've discovered that probably needs to get taken care of:
- [ ] The portrait controller's functions aren't showing up in the LuaBindings
- [ ] I want to be able to do one of these in lua stage.show{character=sherlock, portrait="angry", position="left"} but I'm not sure how to do it. I'm overloading the show function more than I feel comfortable with.

chrisgregan:
Just pushed an option to show inherited members, so you can now inspect the portrait controller members

lealeelu:
Cool! That will be handy.

chrisgregan:
Hmmm.. I seem to have pushed that to a new branch, will sort it out in the morning

chrisgregan:
I pushed that change to your lealeelu:feature/portrait-system-refactor branch. Is there a way to make that change part of this pull request? Happy to just commit it to master either on my end. I'm gonna be honest here - git kinda melts my brain :{

chrisgregan:
btw I noticed an error in the console in the PortratControllerExample scene.
"Unidentified stage position: bored"

I think it's related to your point about overloading the show function. I'd lean towards having different versions of show (e.g. showat, showportrait) or conceptually seperate fucntions altogether rather than overloading it to much.

chrisgregan:
Another option would be to use [named arguments](https://www.lua.org/pil/5.3.html)
These are good when the call signature gets complex or has lots of optional parameters.

lealeelu:
I remember the last pull request I did I thought I saw you were able to commit to my pull request. strange. Maybe I saw that wrong.

Anyways! I could start a new pull request to the branch you have called lealeelu:feature..blah blah
since at this point it feels mostly like cleanup stuff is left to be done.

I would like to use the named arguments. It just feels more intuitive. I'm thinking I would add something like this to fungus.txt:
```
function show (arg)
    return stage.show(arg)
end
```
and use it like this:
```
show{character="sherlock",` portrait="angry", fromposition="right", toposition="left"}
```
Can you just pass args like that? We might need to have a stage lookup in case a stage isn't added to the luabindings...

The showportrait one sounds good too. I'll think about that some more.

chrisgregan:
Ok, that's prob easiest - just include my changes in that new pull request.

You can pass args like that if the receiving function takes a MoonSharp Table as an argument. I'm fine with Fungus having dependencies on stuff in FungusLua (like Table) but I really want to avoid dependencies in the other direction so that FungusLua can always be used independently of the rest of Fungus.

lealeelu:
I was able to get the Moonsharp table to pass values into c# but the table is really hard to work with. Maybe converting it to a list to iterate through would be nice but I can't access this the TableConversions to do it: https://github.com/xanathar/moonsharp/blob/master/src/MoonSharp.Interpreter/Interop/Converters/TableConversions.cs
Is there another way to do it?

chrisgregan:
Is the Table class that hard to work with? You query the DynValues you need from the Table by parameter name and then use the correct accessor to get at the C# value - should just be a few lines of code per property. Not at my laptop to check, but there should be a few examples of this in the FungusLua source.

lealeelu:
I just thought it would be nice to iterate through each supplied value and use a switch statement instead of checking if each possible value exists in the Moonsharp Table. Just being picky. I was actually able to get it to work. It just looks messy. I'll push it up so you can see what I mean.

chrisgregan:
Maybe you could wrap up the gunk in a utility class that takes a show args table and exposes the data in a nice clean way for the rest of the system to work with?

lealeelu:
null checks for everyone! I was using fancier stuff like ?? operators but this straightforward method worked best.

I'll put together some tests.

chrisgregan:
Had a look through yesterday's commit, looks good. What's left on the todo list?

I'm away for a few days in Spain at the minute, but would be good to get this all merged in next week sometime. I fixed some bugs on the plane over so thinking of pushing another beta release next week - would be great to include your stuff.

lealeelu:
Awesome! I think the only things left to do are...

- [ ] Write tests - should have done it before rather than after but I'm still kinda getting used to it.
- [ ] Write Documentation and add to the fungus-site repo

I think I can get those done this weekend. Anything else you can think of?

chrisgregan:
I noticed the portrait controller example was using the show command with the parameters rather than the table version, maybe show how to do it with named args as well? Apologies if that's already done - can't check here!

Would be nice if you could add member comments (just summary is fine) on the new class. I want to start doing a better job of documenting the API myself.

lealeelu:
I think the Portrait Example is done. It has both a show command with named parameters and an example using the new c# commands.

lealeelu:
That PortraitControllerTest isn't passing. I'm getting a strange error message that I haven't been able to figure out. Should I actually put this test in the Lua section instead of Narrative?

chrisgregan:
Lua section might be better yeah. Won't get a chance to look at that error until next week, want to post that message here and I might know what it is?

lealeelu:
The error was because the copy and paste I did between scenes wiped out some of the values along the way. Just had to add some stuff back to the LuaBindings.

chrisgregan:
Ah ok - cool

chrisgregan:
Will start into merging this tonight - good to go on your end?

lealeelu:
Yup! Good to go! Yaaaaaay!

lealeelu:
I also have some documentation in fungus-site

chrisgregan:
Those changes are merged in now #500. Will do the docs next.

PR #505, opened 2016-06-26
Add SimpleScript option

lealeelu:
This is just a start to an idea that I think other people might find useful. It definitely needs some work.

In simplescript, most lines will look like this:
```
r: Hi there!
r/sad/right: I'm feeling sad today.
r/happy: but I have something that always cheers me up!
The destruction of my enemies!
```
You can add lines of lua code by escaping using ``` like you would in markdown.

I took a look at mustache but I couldn't get what I wanted from it. I didn't want to involve my writers in using json.

### DIFF
--- a/Assets/Fungus/Thirdparty/FungusLua/Scripts/LuaUtils.cs
+++ b/Assets/Fungus/Thirdparty/FungusLua/Scripts/LuaUtils.cs
@@ -466,6 +466,36 @@ namespace Fungus
 			}
 			return null;
 		}
+
+		/// <summary>
+		/// 
+		/// </summary>
+		/// <param name="conv"></param>
+		public virtual void Conversation(string conv)
+		{
+			if (string.IsNullOrEmpty(conv)) return;
+
+			char[] delimiters = new char[] { '\r', '\n' };
+			string[] rawLines = conv.Split(delimiters,
+							 StringSplitOptions.RemoveEmptyEntries);
+			
+			//find SimpleScript say strings with portrait options
+			//You can test regex matches here: http://regexstorm.net/tester
+			var sayRegex = new Regex(@"(?<character>\w*)( ?(?<mood>\w*)( ?(?<location>\w*)))?:(?<text>.*)");
+			var sayMatches = sayRegex.Matches(conv);
+
+			foreach (Match match in sayMatches)
+			{
+				string characterkey = match.Groups["character"].Value;
+				string mood = match.Groups["mood"].Value;
+				string location = match.Groups["location"].Value;
+				string text = match.Groups["text"].Value;
+
+				UnityEngine.Debug.Log(text);
+			}
+
+			return;
+		}
    }
 
 }
--- a/Assets/FungusExamples/FungusLua/Narrative/PortaitControlExample.unity
+++ b/Assets/FungusExamples/FungusLua/Narrative/PortaitControlExample.unity
@@ -168,7 +168,6 @@ Transform:
   m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
   m_LocalPosition: {x: 0, y: 0, z: -10}
   m_LocalScale: {x: 1, y: 1, z: 1}
-  m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_Children: []
   m_Father: {fileID: 0}
   m_RootOrder: 0
@@ -185,6 +184,7 @@ GameObject:
   - 114: {fileID: 574033267}
   - 114: {fileID: 574033272}
   - 114: {fileID: 574033266}
+  - 114: {fileID: 574033273}
   m_Layer: 0
   m_Name: Flowchart
   m_TagString: Untagged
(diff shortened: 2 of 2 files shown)


chrisgregan:
I think this would be a great addition, would really speed up the writing process!

From an architecture perspective, I'd like to make this more of a core feature of FungusLua rather than a separate command. If you want to run your script in a LuaScript component you would expect to be able to use the simplescript syntax there too.

Regarding the syntax itself, we should aim to keep the style consistent with Lua, e.g.:

```
s = sherlock
w = watson

-- comments!

s,angry,right: What are we doing here again?
w,annoyed,left: I don't know.
You're the one who called me out here.
s,bored: I can't be bothered to remember every appointment I arrange.
s,angry: That's why I have you.
w: ...

stage.hide(sherlock)
stage.hide(watson)
wait(1)
```

It should be possible to mix simplescript in with Lua freely without special code blocks, the lines with special markup on them just expand into more Lua.

To get this to work generically, I'll add an IPreprocessor interface to LuaEnvironment, and have LuaUtils implement it. LuaEnvironment will run all scripts through the preprocessor before passing them to MoonSharp for execution. That way every script that goes through FungusLua can use this syntax, and we can extend it in the future. 

Sorry for all the proposed changes, but I think this could be a really powerful feature so let's nail it!

What do you think?

chrisgregan:
Just added the [preprocessor support](https://github.com/snozbot/fungus/commit/4c786232d8b4061918c23140d0bd833d84fa825e) for LuaEnvironment / LuaUtils if you want to hook into that.

lealeelu:
Ya! Nail it!

I think I see what I should do with the Preprocessor. I'll have to tinker around with it when I get home.
I wish I was better at regular expressions so I could easily catch the say commands.
Another thing that I was wanting to add in later, but we can wait on it was a scene change. Basically, if I had a line like...

SCENE: train_station

it would turn into...

camera.changeToView(train_station)

Would that be possible?

chrisgregan:
Good stuff. For the scene change thing, you could just make a Lua utility function, e.g.
```
scene(train_station)
```
Same amount of typing, no code changes needed to support it.

Although that said, this gives me an idea. What if the preprocessor always did this:
SAY a,b,c: => say(a, b, c)
SCENE y: => scene(y)

Where SAY and SCENE can be any string and c includes everything up to the next blank line. It'd basically be a generic way to write dialog & other commands without using brackets all the time.

chrisgregan:
Btw I can help with the regex if you like

lealeelu:
Ohhhhh! yeah that would be pretty good having the processing be more generic like that. I'll think about it.

I figured the regex stuff out, just took me a little time to relearn the syntax. The match group stuff is super handy. I used it in the latest commit.

lealeelu:
can I be picky and have it be that if the first \w matches an existing character key in luabindings, it is automatically a say function?

in luabindings:
[ character_watson ]   [ w ]   [ Fungus.Character ]

in the script:
```
w,happy,left: la la la la la la la la la 
```

converts to:
```
say(w, "happy", "left", "la la la la la la la la la")
```

and have a lua function...
```
function say(character, mood, location, text) {
  --set character, stage function, and say function
}
```

lealeelu:
And the second+ arguments would be assumed to be strings.

chrisgregan:
Hmmm... not sure about checking the LuaBindings to infer the meaning of a line of script. Feels like it could be brittle and maybe give confusing error messages when things aren't set up correctly? I'd prefer a simple, clear transformation that only depends on the text in the script, even if it needs a bit more markup. That way it could work for any Lua function call which would be super powerful. 

Ok, here's another idea for a format using a special leading # character to indicate a line that should be transformed. For this to work we'd need to put string parameters in quotes, though I really like the idea of assuming that the last parameter after a colon is a long string (and no quotes needed).

```
-- This transformation would work for any Lua function call
#wait 1
wait(1)

-- You can use a colon before the last parameter to indicate a long string
#say w, "happy", "left": La la la
say(w, "happy", "left", "La la la")

-- Leaving out the #say at the start is interpreted as a say command (special case)
w, "happy", "left": La la la
say(w, "happy", "left", "La la la")
```

Hope you don't mind bouncing ideas for this format around a bit more. Once we implement the format it'll be hard to change so worth getting it right I think!

lealeelu:
I like where this is headed. I don't mind bouncing ideas back and forth. Maybe we should mention this convo on the forums and get other opinions too?

chrisgregan:
Yup, good idea!

lealeelu:
Now I'm kind of back tracking. I don't want to confuse new users to Lua by conflicting syntax. My original intention was to have this be a logic-less language for pure dialog previews. I threw in the extra lua code blocks to be convenient.

lealeelu:
Ok! here's another iteration!

SimpleScript blocks+keep strings (or single quotes?? Does lua do that?)
```
--some standard lua
wait(1)

#say
  w: Hi!
  s,'angry','right': Don't \"Hi!\" me!
#end

wait(2)
#say w: What was that for??
```
Everything in the script blocks are assumed to be scripted says

chrisgregan:
Interesting! Gives me another idea, but let's continue on the forum thread and see what others think.

lealeelu:
Continued on the forums: http://fungusgames.com/forum/#!/fungus-3-beta:simpler-dialog-in-lua

lealeelu:
Moved to #510

PR #507, opened 2016-07-01
Simplify DialogInput to use Unity's InputManager

ideka:
The input(s) used to advance text are now defined by the Submit Button in the current EventSystem's StandaloneInputModule, which is set to just "Submit" by default, defined by default to be the spacebar, enter, and joystick button 0.

Shift functionality has been removed in favor of a Cancel functionality, analogous to the Submit Button used to advance text. "Cancel" is by default defined to be escape, and joystick button 1.


chrisgregan:
Nice improvement, we've had a lot of requests for joystick / controller support so thanks!

PR #508, opened 2016-07-01
Allow for menu options to be picked without clicking.

ideka:
MenuDialog now has a new bool option: `autoSelectFirstButton`. If enabled, the first interactable button added will be selected automatically. This allows for the menu to be navigated without the need of a mouse, with the Vertical or Horizontal axes as defined by the current Standalone Input Module.

Even if this option is disabled, the menu may be navigated without the need of a mouse thanks to a new GameObject added to the MenuDialog prefab: DefaultSelectable. It's just an invisible Selectable that automatically selects itself when enabled. From there, the user can navigate to the first option with the Vertical input axis. The DefaultSelectable is marked as not interactable, so the user cannot navigate back to it.


chrisgregan:
Works great - thanks for the PR!

PR #510, opened 2016-07-03
Create Conversation Function in LuaUtils

lealeelu:
Started a new pull request to start over with our new idea of using luautils.conversation[[ ]]

### DIFF
--- a/Assets/Fungus/Thirdparty/FungusLua/Scripts/ConversationManager.cs
+++ b/Assets/Fungus/Thirdparty/FungusLua/Scripts/ConversationManager.cs
@@ -9,36 +9,24 @@ namespace Fungus
 	public class ConversationManager
 	{
 		protected Stage stage;
+		protected Character[] characters;
 
-		protected CharacterItem[] characters;
+		protected Character currentCharacter;
+		protected Sprite currentPortrait;
+		protected RectTransform currentPosition;
 
-		protected struct CharacterItem
-		{
-			public string name;
-			public string nameText;
-			public Character character;
-
-			public bool CharacterMatch(string matchString)
-			{
-				return name.StartsWith(matchString, true, System.Globalization.CultureInfo.CurrentCulture)
-					|| nameText.StartsWith(matchString, true, System.Globalization.CultureInfo.CurrentCulture);
-			}
-		}
+		protected Character previousCharacter;
+		protected Sprite previousPortrait;
+		protected RectTransform previousPosition;
 
 		protected bool exitSayWait;
 
-		public ConversationManager () {
+		public ConversationManager ()
+		{
 			stage = Stage.activeStages[0];
-		
-			Character[] characterObjects = GameObject.FindObjectsOfType<Fungus.Character>();
-			characters = new CharacterItem[characterObjects.Length];
-
-			for (int i = 0; i < characterObjects.Length; i++)
-			{
-				characters[i].name = characterObjects[i].name;
-				characters[i].nameText = characterObjects[i].nameText;
-				characters[i].character = characterObjects[i];
-			}
+			
+			// cache characters for faster lookup
+			characters = GameObject.FindObjectsOfType<Fungus.Character>();
 		}
 
 		/// <summary>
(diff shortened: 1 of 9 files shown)


lealeelu:
I'm a little stuck. When I use the say dialog, there's no pause and I'm not sure how to add that pause correctly. Also, what's the best way to get the character strings?

chrisgregan:
The third parameter to sayDialog.Say() is waitForInput, set that to true and it should wait for the player to click.

I'd probably make a separate class to manage the parsing and mapping of strings to the characters, portraits and stage locations. These maps should be setup when the scene starts to avoid expensive searching while the game is running. For the characters we could match on the gameobject name or the NameText property. Probably best to support matching on either I think.

Assets/Fungus/Thirdparty/FungusLua/Scripts/ConversationManager.cs
```diff
@@ -0,0 +1,101 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using UnityEngine;
+
+namespace Fungus
+{
+	class ConversationManager : MonoBehaviour
```
chrisgregan: ConversationManager could be a normal c# class, instantiated and initialized from LuaUtils. Would be more encapsulated and less likely to have unexpected side effects than a singleton monobehavior.

Assets/Fungus/Thirdparty/FungusLua/Scripts/ConversationManager.cs
```diff
@@ -0,0 +1,101 @@
...
+				sayDialog.Say(text, true, true, true, false, null, () => { });
+			}
+		}
+
+		public Character CharacterSearch(string search)
+		{
+			Character character = characters.Where(
+				item => item.Key.StartsWith(search, true, System.Globalization.CultureInfo.CurrentCulture))
+				.FirstOrDefault().Value;
+			if (character == null)
```
chrisgregan: I'm trying to avoid using Linq these days as it performs loads of memory allocations which can cause gc issues. I should have been tighter on this stuff from the start myself, but I'm gradually improving systems to avoid unneccessary allocations at runtime (e.g. the recent StringBuilder changes to Writer.cs). No problem leaving it using Linq until you get everything working, but lets do an optimisation pass at the end to tidy up. btw allocations at scene start are not as big a deal, it's the ones at runtime we need to worry about!

lealeelu:
I'm having trouble with the executelua command continuing on to the next command when the conversation[[]] isn't finished yet. Also, it only shows the name of the last person in the conversation and ignores sherlock.

lealeelu:
It works now! What should I do to get rid of Linq?

chrisgregan:
Just thinking there - the 'bool exitSayWait' variable can probably be a local variable, the lambda function should close over it. No big deal either way.

chrisgregan:
To replace Linq you can iterate over the collection and call StartsWith directly on each string. This kinda thing.

```
Character character = null;
for (int i = 0; i < characters.Count; +=i)
{
    if (characters[i].key.StartsWith(search, System.Globalization.CultureInfo.CurrentCulture))
    {
        character = characters[i];
        break;
    }
}
```

Yes, it's completely gross :) 

The benefit of doing it this way is no memory allocations are performed (foreach allocates memory for the iterator). Until Unity ships with a decent garbage collector this stuff is important to avoid triggering GC. I know the rest of Fungus still uses foreach all over the place, I'm working to change that gradually.

Another one is frequent string manipulations should be done using a StringBuilder rather than operating on strings directly. 

Here's the rabbit hole if you want to take a look ;)
http://www.somasim.com/blog/2015/04/csharp-memory-and-performance-tips-for-unity
http://www.gamasutra.com/blogs/WendelinReich/20131109/203841/C_Memory_Management_for_Unity_Developers_part_1_of_3.php

lealeelu:
Oh wow! I didn't know about those problems! I think I've been using foreach a lot :( Thanks for the info. I'll read those links thoroughly.

chrisgregan:
No worries. Unity have been talking about upgrading to a newer version of Mono / .net for years. Later versions have a generational gc which would make this much less of an issue. One day maybe!

lealeelu:
Hows that? Anything else I should do before I rebase into one commit? (make it easier to merge)

lealeelu:
Forgot break! one sec

lealeelu:
Ok that's better. Ready for that merge! Let me know if there's anything else I need to fix.

Assets/Fungus/Thirdparty/FungusLua/Scripts/ConversationManager.cs
```diff
@@ -0,0 +1,112 @@
...
+			SayDialog sayDialog = SayDialog.GetSayDialog();
+
+			if (sayDialog == null)
+			{
+				yield break;
+			}
+
+			//find SimpleScript say strings with portrait options
+			//You can test regex matches here: http://regexstorm.net/tester
+			var sayRegex = new Regex(@"((?<character>\w*)( ?(?<portrait>\w*)( ?(?<position>\w*)))?:)?(?<text>.*\r*\n)");
```
chrisgregan: What happens if there's a space in the name of a character, portrait or position? I see some code in Stage.cs to handle this for positions alright.

Some ideas: 
1. Strings with spaces must be enclosed with quotation marks. Complicates the parser code, but it's an easy to understand rule.
2. Strip spaces from names, or replace with underscores. Easier to implement, maybe a bit harder to understand for users. 
3. Use commas as a delimiter. Split on commas and trim each string. Could be an optional format to support names containing spaces.

lealeelu:
I like the comma separated option. So like, any of these would work:
rinku happy right: blah blah
rinku, sad, offscreen right: blah blah
bad guy, angry offscreen left: muwahahahaha

chrisgregan:
That third case with one comma - seems like there's a bunch of possible permutations there for the last two params? It'd more deterministic if we always insist on comma delimiters when there's a name with spaces.

```
one two three: four
one, two two, three: four
```

chrisgregan:
Oh yeah, another thing. Would it be a good idea to let the user omit any of the parameters, and even mix up the order? Basically, have all three parameters interchangeable and optional. When a character's name is specified that character is set as the default for following commands.

```
rinku sad left: Sad rinku appears on the left
right: Rinku moves to the right
happy: Rinku changes to happy
sherlock angry left: Sherlock appears, rinku stays as she is.
surprised: Sherlock changes to surprised
rinku: Rinku speaks, sherlock stays as he is.
```

Main motivation would be less typing - is it more confusing though?

lealeelu:
That last line I forgot the comma. haha.

Maybe for moving params around without spending too much time on lookups, we should cache the positions and portraits like I did with the characters right when the scene loads? That would be the main thing stopping me from doing that. Or am I worrying too much about it?

chrisgregan:
Yep caching would be a good idea, much faster lookup.

lealeelu:
Still working on it. Just taking my time making sure I'm doing it right.

lealeelu:
I'm not liking where this is going. Is there a better way to do this?

chrisgregan:
In hindsight, caching probably only makes sense for characters and stage positions, as the lookup in the scene for these is very slow. Caching the portraits for a character isn't speeding anything up really, better to just do a simple linear search through the existing portraits array.

The string splitting stuff in Conversation() looks fine to me. Is it SetParams you're not happy with? I think it should work like a priority system, first try to match character name first against any parameter, then try to match portrait against any remaining parameter, then try to match stage position against any remaining parameter. Ignore any unmatched parameters (prob with a warning). I can have a stab at an algorithm for this if you want?

I think GetPortrait() and GetPosition() should be quiet tolerant as well, case insensitive and match with start of string too. 

Anything else you want me to comment on?

lealeelu:
I'm not feeling really confident with SetParams, so yeah, it would really help to have you look over it.

The other weak spot is what I'm digging up in PortraitController. Things aren't working quite as expected. Like, if all I'm doing is moving a character from the right to left, it shouldn't need to do anything but that. I need to make it check for changes before it uses show(options) right away. I changed a little bit in PortraitController to reduce the uneeded alpha changes and movement but I keep on running into problems.

I can make the getportrait and getposition more lenient.

chrisgregan:
Because PortraitController is nice and self contained now (thanks again for that!), it should be possible to enumerate all the possible behaviours and build integration tests for each case. That should root out any problems in there but it'll take a bit of time. 

I'll have some time later in the week to dig into this and the SetParams stuff. Can you give me any more detail on the weirdness in the PortraitController you found?

chrisgregan:
Hey Leah, I'm going to make the modifications we discussed. I'd to add these to your pull request branch. I think if you add me as a contributor on your account I can send you a PR with my changes?

You know git-fu is pretty rubbish so if there's a better way to do this then just let me know!

lealeelu:
It's fine! I just added you.

chrisgregan:
Just went for a nice head clearing run there and then it clicked!

We've got the right functionality, but it's in the wrong place. The SayDialog.Say() method should be where we check for the simple script format. That way you could use the simple script syntax in a regular Say command, or if you're calling SayDialog.Say() directly from script or from a say() function call in Lua. We don't need a separate conversation() function at all, the regular say() function can just support the simple script format via SayDialog.Say().

If you think this is the way to go, can you merge in my changes to your PR and I'll just pull this into master now. I've got a full day tomorrow to work on this so I can look after rejigging it to work from SayDialog.

lealeelu:
That's looking way better! Thanks so much!
I'll try resolving some of these conflicts so it can be merged properly.

lealeelu:
You are github ninja now! hehe :)
I'm going to do some more testing after work. Then give my blessings on the merge. This is going to be a great addition! °˖✧◝(⁰▿⁰)◜✧˖°

chrisgregan:
Cool. I'll work on merging this in tomorrow and rejigging things to work from the SayDialog.

chrisgregan:
I made a few more fixes today and added some integration tests.

I'll keep testing here, let me know if you find any weird behaviour on your end. I'd like to put out a release with this in soon.

PR #512, opened 2016-07-12
Create a Slack-like Chatlog

lealeelu:
Create a new component similar to SayDialog, but looks like a scrolling chat program.


lealeelu:
I think I'll put most of the code into ChatDialog.cs for now and then separate some of the logic into it's own controller once I have something that looks like it's working. For now, I just have some TODO's in there with a plan of how to implement this.

PR #519, opened 2016-08-05
Bugfix for when portrait field is blank

lealeelu:
In the portrait lookup, if you have a blank portrait field on a character, it blows up. This prevents it by checking if there is a portrait to compare against first.


chrisgregan:
Awesome, thanks. I love quick easy to merge fixes like this!

PR #520, opened 2016-08-06
Bail out first thing if comment

lealeelu:
I found that if I had a bunch of commented out lines, the game would hang and eat up a ton of cpu and memory. This fixes that.


chrisgregan:
Nice fix, thanks!

PR #521, opened 2016-08-07
Sync active say dialog when Lua sets the say dialog

lealeelu:
There was a disconnect between Lua's setsaydialog and c#'s SayDialog.activeSayDialog. I did the simplest thing and made a sync function in luaUtils. Maybe not the fanciest solution but it works.


lealeelu:
This is the bit of Lua I tested it on after removing john's custom saydialog and adding the dialog to the luaBindings:

```
setsaydialog(johnsaydialog)

conversation[[
john bored left: Wow, I've got my own SayDialog!
sherlock bored right: Big whoop, so do I.

john hide:
sherlock hide:
]]
say "This say dialog is nice"
```

chrisgregan:
Thanks for the fix Leah!

PR #531, opened 2016-08-30
Fix Regex for character names with "." & "'"

altunsercan:
Old Regex expression did not capture Character names with "." and "'". As a result characters with names like "Mr. Jones" or "Ab'ar" were not registering correctly.

I added . & ' characters into Regex. We may need to add some more characters in future.


chrisgregan:
Thanks Altun!

Issue #477, opened 2016-05-17
Color property for Blocks


chrisgregan:
Add a color property for Blocks to allow to use custom colors.

The color is currently baked into the Block images, so we would either support tinting a standard greyscale sprite (might not look great) or provide a few color options to pick from.

Requested on [the forum](http://fungusgames.com/forum/#!/general:custom-colours-for-blocks)

chrisgregan:
Another request for this on [the forum](http://fungusgames.com/forum/#!/general:flowcharts-feature-request)

zvinless:
Hey Chris, I have a couple ideas for this:

**1. Tinting**
Like you said, use a standard default sprite (probably a white -> really really close to white linear gradient so that it doesn't get too muddy) with a tint color. There could be a color picker in the block's inspector near the "Name" field that sets a property in the Node base class. Default tints can be applied to best match the existing color scheme for backwards compatibility.

**2. Make use of Unity's built in assets that are used in the Animator.**
These can be accessed in GUI.skin.customStyles, although I'm not sure if it's documented anywhere. They tend to look pretty nice and cohesive, especially alongside the animator, but are essentially limited to:

Rectangles: Blue, Teal, Gray, Yellow, Red, Orange
Hexagons:   Blue, Teal, Gray, Yellow, Red (there's only a selected version of the orange image)

Here's how they look compared to the current Fungus theme (ignore the text and arrows):
<img width="568" alt="unitynodes" src="https://cloud.githubusercontent.com/assets/3624751/19831960/d191c6de-9dcb-11e6-9361-5e11950144a3.png">

The color could be chosen per block with an interface similar to this:
<img width="276" alt="colorpicker" src="https://cloud.githubusercontent.com/assets/3624751/19831967/322ca00e-9dcc-11e6-9558-1090b8fe2fb7.png">

The obvious downside to this is that it's not guaranteed to work if Unity changes or removes the flow node assets, and there aren't that many colors to choose from.

**3. Make more color variants of current assets**
Again, like you said, this would potentially look better than tinting but it offers fewer choices and may need to be doubled up to look good for both the light and dark Unity themes.

I think I'm leaning toward the tinting option because it's probably the simplest to implement and by far the most flexible. What do you think?

chrisgregan:
I'd say try option 1 and make the default match the existing colours
reasonably closely. I like the colour swatch picker too so it's easy to
keep your colours consistent but the default colour picker would prob be
[...]

PR #553, opened 2016-10-30
Issue #477: Color property for blocks

zvinless:
I addressed this issue as discussed in the issue notes.

A couple of things:
- I did go ahead and swap the Base64 strings with .png files. I think one reason you may have originally went with the strings is because anything in the Resources folder goes into builds even if it isn't used in game. The file size is negligible, though, and there were already images in the Resources folder anyway. Also, I found that a few of the Base64 strings' textures weren't being used at all, so I did not add them back in as image files.
- The default node colors clearly look different than they used to since the old images had more colors in them. I think the general color scheme still looks right, but I could only get so far tinting the grayscale images.


chrisgregan:
Thanks! I'll review and merge early next week.

chrisgregan:
Nice addition - many thanks! And thanks for a very clean easy to merge PR :)

I think Resources can be place inside an Editor folder to prevent them being copied to builds. I'll make that simple change now.

chrisgregan:
Just did a test standalone build and looks like resources are copied even if they are inside an Editor folder. No worries though, they're very small files.

Issue #225, opened 2015-09-21
Flowchart window usability improvements


RonanPearce:
Drag Select groups of Blocks.
     Group, Ungroup blocks, named groups
     Cut, copy, paste
     Keyboard shortcuts
     Auto scroll to show executing node/selected node
     Double click a node to center it
     Center zoom on middle of screen instead of top left
     Draw a magnifying glass beside the zoom slider
     Zoom buttons to reset zoom to 1:1 or full zoom out
     Improve connector curves rendering algorithm
     Mini map with overview of script
     Add a select block drop down in the Flowchart inspector (when no block selected)
     Dropdown menu in Flowchart window to find a block by name

RonanPearce:
(Comment from Marcin Rsukiewicz)
Maybe you could also add a right-click menu on an empty space in the flowchart window to have a "create block here" command (probably together with "paste here" as well)?

Also the new block button should probably create the block somewhere where the user has centered the flowchart view, rather than at the top/left corner of the view as it does now.

Right now you have to drag the block around to a proper place - and this get annoying quite fast when the flowchart is getting a bit larger, not to mention that when I zoom out fully, the block gets created behind the button I've just clicked.

chrisgregan:
More feature requests in [this forum post](http://fungusgames.com/forum/#!/support-questions:flowcharts-feature-reques).

chrisgregan:
More requests [here](http://fungusgames.com/forum/#!/general:flowchart-editing-suggestio)

zvinless:
I can start on these next week. I'm planning on breaking each sub feature out into its own pull request.

Here's the list of things I'm thinking of tackling:
- Multi-select (left click and drag to make selection box) and drag 
- Keyboard shortcuts
 - Copy Block(s)
 - Paste Block(s)
 - Duplicate Block(s)
 - Delete Block(s)
 - Select All Block(s)
- Add copy option to block's right-click context menu
- Right-click context menu to add/paste block(s)
- Zoom in/out from center **alternatively from cursor position instead?**
- Zoom buttons to reset zoom (and maybe center the graph in case you lose it?)
- Select block drop down in the Flowchart inspector (when no block selected)
- Dropdown menu **or search bar? :)** in Flowchart window to find a block by name

If all goes well I can investigate groups, autoscrolling, and the minimap afterward.

Other questions:
Any preference for right click to center vs. right click to rename block?

Is the "Improve connector curves rendering algorithm" request about performance or appearance? 

[...]

PR #557, opened 2016-11-02
Flowchart window improvements - multi-select

zvinless:
Added multi select feature to flowchart window!  See: #225 

Notes:
- I incremented the flowchart version number and used Flowchart.cs's `UpdateVersion()` function to copy any previously serialized `selectedBock` into the new `selectedBlocks` list for backwards compatibility. The `selectedBlock` field is no longer used at all. If you'd prefer cleaner code to preserving selections from old versions, I'd just keep the flowchart version at 1, delete `selectedBlock`, and ignore the new code in `UpdateVersion()`.
- Before I made these changes, it looked like there was code in place to support undoing block movement, but it wasn't working. I'm hoping this was unintentional, because I hooked it up as part of this feature (see lines 313-315 in FlowchartWindow.cs).
- If you select multiple blocks and drag them around, you might notice that not all of the selected blocks are drawn on top of all unselected blocks (the one you actually click is guaranteed to be, though). This is actually consistent with other graph editors including Unity's Animator, but it bugs me a little bit. Addressing this would probably mean ditching `GUI/Layout.Window()` for direct calls to GUI.Box() and manually managing draw order.

Copy+paste from commit description:
- Added multi-select support via action key/shift+click and click+drag
- Block movement and context menu options now affect all selected blocks
- Flowchart’s SelectedBlock property now refers to the active selected
block rather than the sole selected block. It works similarly to
Unity’s Selection.activeObject
- Bumped flowchart version to 2 to upgrade previously serialized
selected blocks to new format (now a list of blocks rather than a
single block)


chrisgregan:
Works great here! Some minor suggestions before we merge.

I think you could just remove the old selectedBlock field. I don't think anyone will notice if the selected block is no longer selected after an upgrade. No point having old code lingering around unless we have to!

Block movement undo was indeed broken, thanks for the fix.

I wouldn't worry about that draw order issue unless it actually prevents the user from doing something.

Can you change "Multi-block editing not yet supported" to something like "Multiple blocks selected". More of a status message than pointing out a missing feature :)

zvinless:
Went ahead and made those changes. Good call on the status message!

chrisgregan:
Another great feature for Fungus - thanks!

PR #558, opened 2016-11-03
Flowchart window improvements - zoom updates

zvinless:
The flowchart now zooms in and out from either the center of the window or your mouse cursor. See: #225 

Only thing to note is that I'm sorry about the new ugly button to toggle the zoom setting. The toolbar mentioned in the issue will replace it soon :) (wanted to keep it a separate pull request to make things simpler).

### DIFF
--- a/Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
+++ b/Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
@@ -35,10 +35,8 @@ namespace Fungus.EditorUtils
         protected Texture2D addTexture;
 
         protected Rect selectionBox;
-        protected Vector2 startSelectionBoxPosition;
-        protected List<Block> mouseDownSelectionState;
-        
-        protected bool cursorZoom = true;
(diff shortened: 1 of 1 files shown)


chrisgregan:
The centered zoom is working well here.

On the option to zoom from the center or from the mouse position, I'd suggest to just pick one behaviour and stick with it rather than provide a user option. I tried to install PlayMaker / NodeCanvas here to see how they handle this, but the Asset Store seems to be busted right now :(

This article explains why I tend to avoid adding user options unless really needed.
http://www.joelonsoftware.com/items/2006/11/21.html

What do you think?

zvinless:
I mostly agree with that article. I went ahead with zoom to cursor because other graph editors do use it (I checked NodeCanvas) and it's more intuitive to me. The zoom bar still uses the center of the window.

By the way, the Asset Store doesn't seem to work for me until I undock its window; maybe that'd work for you?

chrisgregan:
Yeah you can also go too far with that and start removing features that are actually important - like the headphone jack on the iPhone - it's a balance :)

Asset Store window works when it's undocked, thanks for the tip. And thanks for the contrib!

PR #559, opened 2016-11-03
Flowchart window improvements - toolbar

zvinless:
See: #225 
This replaces the floating UI buttons with a toolbar like in other Unity windows. I added a few buttons that were mentioned in the issue as well (min zoom, max zoom, center).

Notes:
- I personally don't feel like I'd use the min/max zoom buttons ever, but I added them in since someone else requested them. Do you think they add unnecessary clutter?
- The toolbar will get filled with other UI that was requested in the issue, but I didn't want to tangle up the toolbar pull request with each feature that will it for UI. That said, if you don't like how it looks now, I don't want to bank on future updates validating it.

### DIFF
--- a/Assets/Fungus/Scripts/Editor/EditorZoomArea.cs
+++ b/Assets/Fungus/Scripts/Editor/EditorZoomArea.cs
@@ -56,15 +56,16 @@ namespace Fungus.EditorUtils
 
     public class EditorZoomArea
     {
-        private const float kEditorWindowTabHeight = 21.0f;
         private static Matrix4x4 _prevGuiMatrix;
+        private static Vector2 offset = new Vector2(2.0f, 19.0f);
+        public static Vector2 Offset { get { return offset; } set { offset = value; } }
         
         public static Rect Begin(float zoomScale, Rect screenCoordsArea)
         {
             GUI.EndGroup();        // End the group Unity begins automatically for an EditorWindow to clip out the window tab. This allows us to draw outside of the size of the EditorWindow.
             
             Rect clippedArea = screenCoordsArea.ScaleSizeBy(1.0f / zoomScale, screenCoordsArea.TopLeft());
-            clippedArea.y += kEditorWindowTabHeight;
+            clippedArea.position += offset;
             GUI.BeginGroup(clippedArea);
             
             _prevGuiMatrix = GUI.matrix;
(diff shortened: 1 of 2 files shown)


chrisgregan:
Wow - looks great! I agree the min / max buttons are redundant as its so easy to zoom in / out anyway. Once we have the full set of functionality in the toolbar we could maybe look at making icons for some of these options.

That reflection stuff to check isDocked looks a little hairy, but I guess it's necessary? Only real concern would be that it works ok in Unity 5.0, but I can check that myself later before we release.

oh one other thing. How does the toolbar look in light vs dark skin? I'm on Unity Personal here so can't check it easily.

You are going to make a lot of users very happy with these improvements ;)

zvinless:
I'm glad you like it! Fungus is an awesome tool so I'm glad to contribute :).

The reflection stuff is because the `EditorZoomArea` ends up looking a little off depending on if the window is docked or undocked, and the toolbar made it way more noticeable (you can actually see related issues in other graph editor plugins). The reflection code was the only way I could figure out how to check if the window is actually docked or not. I have been testing back and forth between the initial Unity 5 release and 5.4.2 before committing anything, so it should be good for current releases (the offsets also look correct in Unity 5.0), but there's always a chance of things changing in the future. If we knew, for instance, that nobody is using that window undocked, we could force the docked offsets and stick to those. For now, I can update that code to at least do nothing if it fails rather than potentially cause an error?

The toolbar uses `EditorStyles` which automatically use the appropriate skin for the theme, i.e. it's for the pro theme it's light text on darker backgrounds just like the toolbars in other Unity windows. It looks like there's only one "+" icon, though, which is too dark to see well using the dark theme. From what I can tell a pro theme variant may have been used before the Resources/Icons folder was added?

I'll go ahead and update the reflection code for future safety--do you want me to remove those toolbar buttons too or just leave them until there's more stuff in the bar?

chrisgregan:
I would think people with a second monitor would likely use the Flowchart window undocked. If it's working ok in 5.0 and 5.4.2 then that's fine. If it breaks in future we'll no doubt get bug reports about it. 

I made light and dark versions of the old resources yeah. We'd need similar for the new ones in Resources/Icons and code to select between them. Are you able to check the dark skin there?

I'd say remove the Min / Max buttons now if we're not planning to have them in the next release. Even with just the existing controls it looks a lot better to have the toolbar.

zvinless:
Alright, removed those buttons!

After thinking about it more, to avoid dead code I'd rather just have the reflection code fail (not silently) if something about editor windows changes in the future. Shouldn't be too hard to deal with a bug report about it.

And yeah, I can check dark skin. If you want I can look into getting the old dark skin icons into the icons folder and hooking them up next?

chrisgregan:
Good call, no way we'd miss it that way. Dark skin icons would be great.

I'm at a conference today, will merge the PR when I get a chance.

chrisgregan:
Merged the PR, thanks!

PR #560, opened 2016-11-07
Flowchart window improvements - keyboard shortcuts and context menu updates

zvinless:
See #225

Added duplicate, delete, and select all commands. Also added relevant commands to the block context menu and added new context menu with an add new block option.

In order to support a right-click context menu when clicking on the background while still being able to drag and pan, I moved showing the context menu from `EventType.ContextClick` to `EventType.MouseUp`, because `EventType.ContextClick` occurs on mouse down (at least on macOS). Unfortunately, there is some magic that happens before or during the `ContextClick` event, so the new context menus have slightly smaller text than before. I'm not really sure how to change this, but it seems fine regardless.

By the way--I did do the pro icons in another branch as discussed, but I want to try out one more idea before I do a PR with them.

### DIFF
--- a/Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
+++ b/Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
@@ -39,8 +39,6 @@ namespace Fungus.EditorUtils
         protected Vector2 startSelectionBoxPosition = -Vector2.one;
         protected List<Block> mouseDownSelectionState = new List<Block>();
 
-        protected List<Block> copyList = new List<Block>();
-
         // Context Click occurs on MouseDown which interferes with panning
         // Track right click positions manually to show menus on MouseUp
         protected Vector2 rightClickDown = -Vector2.one;
@@ -67,7 +65,6 @@ namespace Fungus.EditorUtils
             nodeStyle.wordWrap = true;
 
             addTexture = Resources.Load("Icons/add_small") as Texture2D;
-            copyList.Clear();
         }
 
         protected virtual void OnInspectorUpdate()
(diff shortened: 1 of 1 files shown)


zvinless:
Ah, crap, just found a bug with cutting and pasting. Wait to merge this...I'll try to fix it as soon as I can.

chrisgregan:
No worries!

zvinless:
Okay, I removed all the hard stuff to to leave it for a future PR :P. I think what's left is still useful enough to warrant a merge. I'll update the description to reflect the changes.

Basically, the helper functionality in Unity to copy components (`ComponentUtility.CopyComponent() / PasteComponentAsNew()` and `EditorUtility.CopySerialized()`) work well for duplication, but they work on one component at a time, so to copy multiple components I need to copy and paste everything to an intermediate clipboard first. That registers undos (inherent in the utility function calls), and I don't want to be able to undo my copy (only the paste). So I'll need to deal with copying components to a clipboard without undos, which I think could take a while. I'm planning on finishing up other feature branches and coming back to this for a part 2.

chrisgregan:
Just tried it out. The context menu seems to only appear intermittently? You've probably seen it already, but if you look at BlockEditor.Copy() you can see how we do it for the command list context menu using a "CommandCopyBuffer" gameobject.

I'll hold off merging so until you get back to look at this one.

zvinless:
Oh, wow, I actually didn't see that! I'm glad you pointed it out, because as one of my attempts I made the exact same thing but for blocks. It does have the same issue, though, in that you can undo copies, which I want to try to avoid (let me know if that seems strange to you). I quickly tried out using a buffer of SerializedObjects and so far so good, but I want to test it really thoroughly.

About the context menu--I'm thinking it might be because your right click mouse down and up positions have to be identical for it to appear. I'll try increasing the tolerance a bit. Also, out of curiosity, which OS do you primarily use? I'm testing everything on Mac versions of Unity but I'm thinking I should use Bootcamp and make sure it behaves well on Windows.

chrisgregan:
Context menu works fine after the tolerance change. I'm running Unity on OSX here with a Magic Mouse (with the touch sensitive casing), might have been causing some jiggling.

SerializedObjects sounds like a promising approach. Another idea would be to use Undo.Clear to wipe the undo stack altogether. Obviously not an ideal option, but robust at least. https://docs.unity3d.com/ScriptReference/Undo.ClearUndo.html

Should I keep this PR open or merge what's there now?

zvinless:
As long as everything in this PR is working well I'd go ahead and merge it. I'll probably need to wait until the weekend to do much more with copying and pasting, so I'll open a separate one for that. I'd like to see what can be done with `ClearUndo()`.

chrisgregan:
Merged now - thanks.

PR #561, opened 2016-11-10
Pro skin icons and editor resources restructure

zvinless:
I added pro skin icons and also changed some things about how the editor textures are loaded.

This was kind of an experiment I did for fun but I think it worked out okay. I had a few goals:
- Get editor textures out of builds
- Unify how editor textures are loaded
- Ensure that editor textures still load property if folders get moved around (to an extent)
- Ensure that users won't cause collisions with the editor textures (e.g. file with identical name in different Resources folder)
- Minimize amount of coding necessary for editor texture updates

I reworked `FungusEditorResources.cs` into a partial class; the new part of it is auto-generated code based on textures found in the new "Fungus Editor Resources" folder. There's a new menu option in Tools -> Fungus -> Utilities to update this script if textures are added or removed during Fungus development. Otherwise, texture loading hasn't changed drastically. Pro icons need to go into a folder called "pro" (case-insensitive) anywhere within Fungus Editor Resources and have identical names to their free skin counterparts.

Also, I removed the resize handle and left and right arrow icons because I couldn't find any references to them.


chrisgregan:
Interesting way to do it with the code generator. I'd have been fine with hardcoding it as these resources are probably not likely to change very often, but the code generator approach is perfectly valid. It's a clever way to avoid including the editor resources in builds.

The only change I'd suggest is renaming "Fungus Editor Resources" to "EditorResources". We don't use spaces in any other folder in Fungus so lets keep that consistent, and the "Fungus" part is redundant as it's already inside the Fungus folder. 

Yeah sorry, I am very picky about file & folder names :)

zvinless:
Haha, I was actually going to say feel free to nitpick the folder naming. The reason I added "Fungus" in front of it is to reduce the chances of somebody else making a folder with the same name somewhere else in their project and causing a collision with one or more of the textures. It's probably overkill to try to not dictate a folder structure WITHIN the Fungus folder, but oh well...

Speaking of overkill, I did go a step further locally and made an auto-generated `ScriptableObject` singleton (as singletony as `ScriptableObject`s can get, at least) asset with references to each editor texture. That way you only have to search for that one asset to find all the textures, and it's safer because the references are on a GUID level rather than by name. I felt like it was maybe a bit much and committed the simpler static-only version, but if you're interested let me know...

Otherwise I can go ahead and just change the folder name. I don't think it's too likely that there will be collisions regardless :).

chrisgregan:
It's up to you, if you think the ScriptableObject solution is an improvement then feel free to send it in. How does a singleton asset work btw? Is it a static method that returns a reference to the asset?

Also, I'm looking at the icons and I think they're a little blurrier than they used to be?
before
after

zvinless:
The top image definitely looks blurrier, but I think it's just the box outlines and not the actual icons? I've been switching back and forth between branches and my icons always look like the bottom image regardless of how they're getting loaded. Do you think there might be any other step to reproduce that?

And yeah, I'm not sure if it's even proper to call an asset a singleton because it's a lot more difficult to enforce that one exists at any given time. Basically the static property will find the asset and load it as a static reference, and possibly delete duplicates. To be really strict it's probably best to add an AssetImporter as well to do something about duplicating the asset outside of code, and somehow stop `ScriptableObject.CreateInstance()` from working, but realistically if it's an internal editor class you wouldn't have to worry about users unintentionally making new instances.

chrisgregan:
Actually, I think the blurring is unrelated to your changes. I took the screenshot of the sharper version from a version a few releases back. I'll bet it's caused by the changes I made to support retina display - sorry for the red herring.

Yeah I think it's fine not to worry about all those singleton restrictions. It's an internal editor class that users aren't expected to interact with. No point!

PR #563, opened 2016-11-12
Flowchart window improvements - connector curve updates

zvinless:
See: #225

This one should be pretty straightforward. I replaced all of the GLDraw code with built-in Unity stuff, mostly using the Handles class. I also swapped out the background to use the same texture as the Animator Controller window. On the free skin, it looks pretty much the same as it used to except there's the drop shadow around the perimeter. On the pro skin, it's darker than it used to be, but now it matches other Unity windows.

Let me know if you'd change anything, e.g. line weight and lightness of connection curves or the texture or size of connection points (they're just flat blue circles for now).


chrisgregan:
Another nice improvement - thanks.

chrisgregan:
Just as a thought, I believe Node Canvas has an option to draw the links either as curves or straight lines. Do you think uses would appreciate this option?"
screen shot 2016-11-14 at 13 19 07

zvinless:
I saw that when evaluating Node Canvas, too, but kinda feel like it's one of those too many unnecessary choices things. That said, it'd be quick to implement; we'd just want to offset the lines by some radius so that opposite lines between blocks don't overlap each other (like the Unity Animator Controller graph). I'll look into after I get through another couple pull requests :)

PR #566, opened 2016-11-14
Portrait location change to inline or stage

lealeelu:
In the conversation system, we want to add the ability to move the portrait from the stage to inline on the say dialog and back again using the "inline" and "stage" option params.

### DIFF
--- a/Assets/Fungus/Scripts/Components/Character.cs
+++ b/Assets/Fungus/Scripts/Components/Character.cs
@@ -29,6 +29,9 @@ namespace Fungus
         [Tooltip("Direction that portrait sprites face.")]
         [SerializeField] protected FacingDirection portraitsFace;
 
+        [Tooltip("Show the character in the say dialog or on the stage.")]
+        [SerializeField] protected PortraitLocation portraitLocation;
+
         [Tooltip("Sets the active Say dialog with a reference to a Say Dialog object in the scene. This Say Dialog will be used whenever the character speaks.")]
         [SerializeField] protected SayDialog setSayDialog;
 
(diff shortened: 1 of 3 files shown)


lealeelu:
Added the tests. Still working on making them pass.

chrisgregan:
Ok cool, good luck!

lealeelu:
I have the bare-bones portrait switching working. I want to add some more tests to check when there are multiple characters on the stage that switch back and forth.
@chrisgregan , are there any other strange cases you can think of that I should test?

chrisgregan:
Is this ready to merge now or have you more to do on it?

lealeelu:
I'm going to look over it a bit more.

lealeelu:
I'm rethinking how we should do this. What if we treat "inline" as if it were just another stage position like "left", "right" and "middle". So instead of....
john right: on stage right
john inline: now I'm in the dialog box
john stage right: back on stage right

john right: on stage right
john inline: now I'm in the dialog box
john right: back on stage right

Although, would we get weird transitions going from stage to inline as if it were trying to animate on stage?

chrisgregan:
I like this option, it's much cleaner than adding a whole new param type. 

The conversation manager would need to be aware of the special 'inline' stage position and correctly handle switching between the two modes. To be safe, we could log an error message if the stage contains a position with the name 'inline'.

lealeelu:
Sounds good to me! I'll fix that up.

lealeelu:
Going to start a new branch.

andyandy2015:
Thanks a lot for this. Question: Does it work when I have custom dialog box per each character? Thanks.

lealeelu:
It should as long as the custom dialog still has the portrait game object to put it into.

andyandy2015:
Ok. Thanks. I will try

PR #567, opened 2016-11-15
Flowchart window improvements - cut/copy/paste

zvinless:
See: #225 
Finished implementing copy and paste using `SerializedObject`s. I think it ends up being the simplest solution because there's no need for a hidden GameObject and otherwise it's about the same amount of code as using reflection. You can also copy and paste between flowcharts, which maybe someone will find convenient. I went ahead and replaced the `DuplicateBlocks` functionality with a copy/paste combo to keep the functionality all in one place, but I honestly didn't profile it so see if it's slower or anything.


Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
```diff
@@ -8,11 +8,94 @@
...
+            }
+        }
+
+        protected class BlockCopy
+        {
+            private SerializedObject block = null;
+            private List<ClipboardObject> commands = new List<ClipboardObject>();
+            private ClipboardObject eventHandler = null;
+
+            internal BlockCopy(Block block)
```
chrisgregan: Just curious - is there a reason this is marked internal?

chrisgregan:
Works great here. Small question above about the internal keyword.

Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
zvinless: I think it's the most restricted scope while still being accessible elsewhere in the outer class? It could just as well be public; wouldn't make any actual difference. Want me to change it?

chrisgregan: Naw, it's fine - was just wondering. Merging now.

zvinless: Alright, I mean I'm by no means a C# guru so if something seems bizarre please let me know! I appreciate it. But I also hope you don't have to spend tons of time reviewing all my commits 😄 .

PR #568, opened 2016-11-17
Editor resource loading modifications

zvinless:
This is the other method of getting editor textures that I mentioned earlier. This is definitely not the most valuable thing to put effort into, but I've been curious about the very best way to do it. The general process is simpler (find the one asset and then use its references instead of finding all the individual textures) and there's no possibility of name collisions, but there's potential for other complications since there's a new asset that could get accidentally modified if someone is poking around. Also, removed the menu option from before and moved it into a button on the asset's inspector, so it's less user-facing since it's only really intended for development purposes.

I *think* it's a better implementation, but I doubt anyone would notice any difference unless there are bugs or name collisions, etc.


chrisgregan:
Yep, that's a better way to do it. I like the way the code generator button is now part of the scriptable object instead of the main menu as very few users will need to know about it.

I'm a big fan of refactoring so PRs like this that just improve the implementation of existing functionality are great.

PR #573, opened 2016-11-24
Flowchart window improvements - drop down search filter

zvinless:
Added the search filter basically how you described in the comment in #225. I actually had the same idea, but it was giving me trouble using a `PopupWindow`, so I ended up just drawing a box after the other controls. 

Controls:
- Click on search bar or cmd+f/Find to start searching
- Up and down arrow keys move selection, which auto centers to selected block
- Click on block or press return to center and select it
- Escape key or clicking outside of the popup exits it without selecting anything

There are some details that could be adjusted, like the point in which a block gets selected, the ordering of blocks (alphabetize?), whether or not text gets cleared when the popup disappears, etc., and it's not too hard to see how this could be expanded to support multi-selection, but I feel like it's fine where it's at.


chrisgregan:
Wow - that works spectacularly well. It's so fast to a specific block now, great job!

Of the suggestions, maybe alphabetizing the list would be nice but don't think it makes much difference really. Multi select sounds kinda overkill to me, but if it can be done without complicating the interface then no harm I guess.

PR #577, opened 2016-12-04
Flowchart Window refactor

zvinless:
Still working on this, but I'm basically just trying to clean some code up and improve the performance of the flowchart window.

**Refactoring:**
I wanted to make the event handling code less all over the place, but that's apparently pretty tough to do with the IMGUI system. The flow is now basically like this:
1. Pre-process events that built-in GUI controls will use (eat). This seems to be necessary if you want to attach any functionality to controls. An example would be clicking on the search bar.
2. Draw all of the main content.
3. Draw all of the built-in GUI controls, e.g. the toolbar, (after main content so it's layered above it).
4. Handle events for content drawn in part 2. This allows the built-in controls to handle the events first, which is more intuitive since they're layered above the main content.

This didn't work totally seamlessly, but I think it's a little easier for me to follow than how things were previously.

**Performance:**
After profiling the window a bit it became obvious that garbage was the biggest issue. On one of my computers, dragging lots of blocks around, for instance, is really choppy. I've tried to reduce allocations and also get things out of OnGUI if they can be handled on more per-event basis. On my Macbook Pro, I can now use 100+ block flowcharts without it getting choppy. I regret doing this all in a single PR, though, because I'm not sure which changes have made the biggest impact and I'd rather not make them at all if they aren't helping much. Let me know if you'd prefer to integrate things more gradually (or at all). I think the next significant thing to look at would be the variables window, which can make the window start to chug pretty easily.

I'll ask questions in the code.


Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
```diff
@@ -531,129 +583,343 @@ protected virtual void DrawBlockPopup(Flowchart flowchart)
...
+            Rect scriptViewRect = new Rect(0, 0, this.position.width / flowchart.Zoom, this.position.height / flowchart.Zoom);
 
-            // Draw connections
-            foreach (var block in blocks)
-            {
-                DrawConnections(flowchart, block, false);
-            }
-            foreach (var block in blocks)
-            {
-                DrawConnections(flowchart, block, true);
```
zvinless: Do you remember why this function was called twice? The second call obviously draws the highlighted versions of the connections; do you think it was just to ensure that highlighted connections were layered on top of the others?

Assets/Fungus/Scripts/Editor/FlowchartEditor.cs
```diff
@@ -82,14 +82,6 @@ public override void OnInspectorGUI()
                 EditorWindow.GetWindow(typeof(FlowchartWindow), false, "Flowchart");
             }
 
-            GUILayout.Space(10);
-
-            if (GUILayout.Button(new GUIContent("Center View", "Centers the window view at the center of all blocks in the Flowchart")))
```
zvinless: This was actually kind of broken after I modified the zooming and centering. I figured it wasn't relevant unless the flowchart window was actually visible anyway (and now there's a center button in the window), so I got rid of it from the inspector.

Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
chrisgregan: Yep, that was it. It looked weird when a non-highlighted block overlapped in front of a highlighted one. It seemed easier to do it this way rather than maintain two separate lists for highlighted / non-highlighted.

Assets/Fungus/Scripts/Editor/FlowchartEditor.cs
chrisgregan: Yes, this button is now redundant so no problem removing it.

Assets/Fungus/Scripts/Components/Flowchart.cs
```diff
@@ -319,18 +316,13 @@ public static void BroadcastFungusMessage(string messageName)
...
-
-        /// <summary>
         /// Current actively selected block in the Flowchart editor.
         /// </summary>
         public virtual Block SelectedBlock
         { 
             get
             {
-                return selectedBlocks.FirstOrDefault();
+                return selectedBlocks.Count > 0 ? selectedBlocks[0] : null;
```
chrisgregan: Is this not the same thing?

chrisgregan:
The changes look great so far. Nice idea to make the event handling more structured.

btw let me know if you have any ideas to improve the layout of the variables window. I tried a few different options and ended up settling on the floating window approach, but still not convinced it's the best way to do it. It might work better as a collapsible side panel for example. I think it does need to be part of the Flowchart window though so it's easy to see the values changing at runtime.

Assets/Fungus/Scripts/Components/Flowchart.cs
zvinless: I believe I did this because the profiler said `FirstOrDefault()` had a > 0 CG Alloc. I have no idea why it would, though. I'll double check.

chrisgregan:
Hey - is there more work to do on this or can I merge it now? Getting ready for the next release once I clear out some bugs this week.

zvinless:
Hey! Sorry I haven't updated this in a while--been way busier than I realized I'd be. There were still some things I wanted to look at in this PR, so I'd say leave this out of the next release. I would like to get to this afterward, though. Hopefully I can get back to it soon!

chrisgregan:
No problem, I'll hold off merging for now.

zvinless:
I've been looking at this a bit today and I think I actually want to close this PR so that I can split it up into separate, smaller PRs. I want to spend more time benchmarking the optimizations so that I don't do anything that isn't worthwhile that might introduce bugs.

PR #582, opened 2017-01-09
Refactor EventHandler notification in order to avoid FindObjectsOfType calls

sp-francisco-ruiz:
FindObjectsOfType is an expensive call and it is being called for each event in Draggable2D and Clickable2D. 

In order to avoid them I considered 2 options:
 - Register and unregister the Handlers to the desired object, so the object would notify directly to the EventHanldlers. I've implemented this when the object needs to access the Handler.

 - Register the handlers to an EventDispatcher and raise events from the object. I decided to implement this because with the EventDispatcher we can make it more isolated, only registered objects would listen to events, and multiple objects can listen to a certain event avoiding multiple references in objects. Also this way of dispatching events can be reused in multiple places.

### DIFF
--- a/Assets/Fungus/Scripts/Components/Clickable2D.cs
+++ b/Assets/Fungus/Scripts/Components/Clickable2D.cs
@@ -39,13 +39,7 @@ namespace Fungus
                 return;
             }
 
-            // TODO: Cache these objects for faster lookup
-            var handlers = GameObject.FindObjectsOfType<ObjectClicked>();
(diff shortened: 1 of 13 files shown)


chrisgregan:
Looks like a great improvement! Will try to review and merge this week.

chrisgregan:
Ok looks good. There's a few style changes I want to make, but I'll just do that myself after merging. Thanks!

PR #584, opened 2017-01-15
Removed "Center View" button from FlowchartEditor

zvinless:
Same as the previous pull request. I figured this might be good to do before another release since the button didn't work correctly any more.


chrisgregan:
Thanks!

PR #588, opened 2017-01-23
Flowchart window event refactor

zvinless:
This is a lot of the same code from the other pull request, except I tried to avoid doing any optimizations. It's mostly just refactoring how the events are handled and also some misc. cleanup, e.g. keeping the flowchart as a class variable. It's still a bigger PR than I'd hoped. It's hard to isolate stuff in editor code :(.

### DIFF
--- a/Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
+++ b/Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
@@ -11,7 +11,7 @@ using Object = UnityEngine.Object;
 
 namespace Fungus.EditorUtils
 {
-    public class FlowchartWindow : EditorWindow
+    public class FlowchartWindow : EventWindow
     {
         protected class ClipboardObject
         {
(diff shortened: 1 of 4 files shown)


Assets/Materials.meta
```diff
@@ -0,0 +1,9 @@
+fileFormatVersion: 2
```
zvinless: This meta file kept appearing for me, so I'm guessing it wasn't added previously?

Assets/Fungus/Scripts/Editor/FlowchartWindow.cs
```diff
@@ -1205,7 +1125,7 @@ protected virtual void DrawConnections(Flowchart flowchart, Block block, bool hi
 
             var connectedBlocks = new List<Block>();
 
-            bool blockIsSelected = (flowchart.SelectedBlock != block);
+            bool blockIsSelected = flowchart.SelectedBlock == block;
```
zvinless: This was making it so no connections were highlighted. I think it's been like that for a while. Seems fine after the change.

chrisgregan:
Looks good, thanks for the improvements

PR #590, opened 2017-02-01
fix portrait hide bug in conversation

lealeelu:
Fix I discussed in Gitter chat.
If you hid a portrait twice, it would show up on the second hide. Now you have to specify a portrait to show before it will appear on stage again.


lealeelu:
This should be good to merge whenever you're ready. I've been using this change on my game and so far so good.
Thanks!

chrisgregan:
Thanks for the bug fix Leah!

PR #592, opened 2017-02-10
Stops Conversation Audio

### LINKED ISSUE #591: Custom Beeps keep playing into next dialog in Conversation
I added a custom character sound effect to character 1. Then I made character 2 to act as the narrator with a custom narrator dialog box that is muted. When I have a dialog from character 1 and then have the silent narrator dialog, the sound effect from character 1 will play during character 2's dialog for a couple seconds. 
Here's an example of what I'm talking about: https://youtu.be/05CbzRkZn7s

It only seems to happen in the Conversation system. I tried to recreate the bug with two say dialogs and it had no problem.

lealeelu:
fixes #591 
I wasn't sure how to add a test for this bug, but it might not be necessary...


chrisgregan:
Looks good thanks!

PR #593, opened 2017-02-11
Dialog history

lealeelu:
Super rudimentary history display/manager. Needs way more work.


lealeelu:
Got stuck on the history loading more than it should or the restart not resetting the text in the history view.

chrisgregan:
Well the approach looks good :)

Debugging this stuff can be tricky. It might help to dump the Save Point data to a file so you can check exactly what's being saved for each save point. For the restart issue, I'd suggest adding a new Restart signal to SaveManagerSignals.cs which gets sent when the Restart button is pressed (i.e. in SaveMenu.Restart). That way the HistoryManager can clear itself when the game restarts.

btw I think something like NarrativeLog might be more descriptive than HistoryManager (which could be confused with the SaveHistory).

lealeelu:
I want to get that reset signal working so we don't have to be constantly updating the text in the history view. Maybe do the same for the Debug view and take it out of the savemenu.update?

lealeelu:
Ok, I'll walk away for a bit and let you take a look at what I have so far.
I tried to separate out the narrative log from the save menu as best I could for now so that people have the choice of having it or not. It would be nice to be able to disable and enable each menu button depending on what people need. Like having a generic Menu script that adds/removes the savemenu, narrativelog, or whatever else we might add. 
I myself would like a general menu that I can change text speed, bgm volume, and beep audio.
THOSE SIGNALS ARE AWESOME!

chrisgregan:
Works great, well done! Good idea to separate the Narrative Log Menu component out. You could add a boolean property on that component to control if the Narrative Menu is available. Might be useful for people using Fungus for games that don't have much/any dialog text but still want to save.

Not sure you need the Log button actually - could just show the Log whenever the Save Menu is expanded? One less button to fiddle with! The log really helps to visualize the rewinding too.

For button configurations, I think you could just disable any buttons you don't need in the Save Menu? Not much benefit to splitting the standard save options out into separate components.

I like the idea of adding volume controls, etc. to the Save Menu. Seems like a natural place to put them - maybe it could be more of a toolbar stretching across the full width of the screen with all the controls laid out horizontally?

lealeelu:
Having the option to disable to Narrative Log absolutely should be an option. I'll add that in.

As for expanding the menu across the screen, could we make a modal window instead? My goal is to eventually do a lot of mobile game development and Vertical screens seem to be the new hotness. I won't have much room across the top if that's the case. It would be nice to have a single menu button that could take us to the log, save options, sound options, etc.. Not 100% on that yet though.

I've been looking [here for inspiration](http://visual-novel-interfaces.tumblr.com/) but I need to take some time and look for more visual novel interface examples with vertical layouts.

lealeelu:
Ok, so after more research, Vertical screens might not be big in VN's yet, but I'm gonna make it big! Nintendo has been pushing for vertical layout and I think I prefer it. Also, they said it was cool on Cool Games Inc and I take them very seriously hehe. Warning, some cursing: https://soundcloud.com/coolgamesinc/episode-42-the-lodge#t=6:30

I can always move the components myself, so I won't worry too much about making it look good for vertical layout. I'll make sure it looks good in horizontal first. If the layout is responsive, we don't have to worry!

chrisgregan:
I guess the script components don't care about the layout, so if it made sense you could have portrait and landscape versions of the prefab rather than trying to adjust the layout from code.

lealeelu:
so here's what else needs to be done that I'll be working through:
- [ ] Write tests
- [ ] Remove separate narrative log button (people will be able to add this since I'll keep the original method to show it)
- [ ] Add option to disable narrative log
- [ ] Add option to show previous lines or currently showing line - I'll use the previous lines as the default option
- [ ]  Keep the save menu and narrative log on the same component
- [ ] Add some sliding LeanTween action for flare

chrisgregan:
Nice! ...save menu and narrative log on the same gameobject or same component?

lealeelu:
Woops meant gameobject - so on the same save menu UI Prefab

lealeelu:
Where should I put the narrative log tests? I wanted to put it with some existing save menu tests.

chrisgregan:
Heya - prob going to push a new release soon. Is this near ready for merge?

lealeelu:
It's probably two commits away from finished. Still need to fix the scrolling and get the text from a different spot. I'll take a look at that tonight and try to get it add in to the pull.

chrisgregan:
Cool cool!

lealeelu:
I can't get the padding on the text to work! GAR! That's the last thing I needed. Do you have time to take a look at it? I'll keep poking around this evening but I've already spent too much time getting the darn thing to work. ;____;

chrisgregan:
Got it running here - looks grand. What's the padding issue? 

There is a weird thing were the scene is always in a modified state. When I try to save it, it immediately changes to an unsaved state. Is there a value being set every frame in the editor or something?

lealeelu:
That's one of the issues. You can't have a layout element and a Content size fitter on the same object - which I put on Content. It makes it work, but it throws a fit and will never show it being saved. Disable either one of those, and after the next time you stop playing, the interface gets weird.

lealeelu:
Also, I'm working on a new example and forgot to get rid of the old say history example.

lealeelu:
Ok! I think I have it figured out! I just put that all in a new vertical layout panel. It should be good now. Merge if you think it's good to go.

lealeelu:
One quick thing, can we change the menu to a bars icon since it's more than just settings and saving? http://fontawesome.io/icon/bars/

lealeelu:
I think there is a problem when you introduce more than one say dialog. I have a narrative, normal say, and a digital style say window and it keeps on getting stuck on the narrative dialog. Maybe it's the timing on when it gets the activeSayDialog.

lealeelu:
I think I know what the problem is. The GetActiveSayDailog() doesn't get the dialog being used when it's attached to the character. It sees 3 or 4 different active dialogs and defaults to the first one in the array and doesn't change from that.

chrisgregan:
Hey Leah. Yeah no problem switching to bars if you think it makes more sense, it's easy for folks to customise anyway.

So when we're using a SayDialog from a character it should also set the current active Say Dialog?

lealeelu:
I haven't looked too much into it yet. I have for a few of my characters, a custom dialog box attached for them. I would think it should set the active say dialog to that, because that's where I'm getting the text from. Otherwise I get a bunch of "NameText: Story Text" lines in my narrative log.

chrisgregan:
Ah - are you using the conversation command to display the text? ConversationManager.GetSayDialog() should probably set the SayDialog.ActiveSayDialog property when that happens. Sorta weird for a getter to have side effects like that but I think it makes sense?

lealeelu:
I'll add the new icon and see about updating the ConversationManager.GetSayDialog() tonight.

lealeelu:
It's working with my game now, which with all the crazy stuff I do with that, means that it's done for me. Anything else you want to put in this pull request?

lealeelu:
or maybe a rebase?

chrisgregan:
The functionality is working great, but seems the text box layout is off to the left for some reason? Also the button texture doesn't appear for me?

screen shot 2017-03-14 at 17 35 11

lealeelu:
There was some weirdness with the commit last time I guess. I was pretty sick when I was trying to finish it lol
Give it another go.

lealeelu:
Ran into this problem with the content text component: https://forum.unity3d.com/threads/ui-text-character-limit.359729/ The thread has some possible fixes at the end of it.

lealeelu:
I plan on doing something like this in the second pass: https://github.com/qiankanglai/LoopScrollRect

chrisgregan:
oh that's a pain, I would have thought Unity UI would handle that under the hood - it's a common use case. Anyway, the LoopScrollRect approach sounds like the right way to solve it.

lealeelu:
I won't be able to work on this for a couple weeks. I have some big deadlines coming up. If you make a new branch in your repository, just for these menu changes, I can change the pull request to point to that instead of master. It would be nice to start with a new pull request to add in the LoopScrollRect.

chrisgregan:
Donezo!
https://github.com/snozbot/fungus/tree/dialog-history

lealeelu:
Changed to the dialog history branch. Should be good to merge into there.

chrisgregan:
Hey - do I need to do something here? Started a new job recently so been a bit all over the place :)

chrisgregan:
Is it just a Rebase and Merge?

lealeelu:
Congrats on the job! All you should have to do is click the merge button. It'll be put into the dialog-history branch, leaving the master branch alone.

chrisgregan:
Ok, that's done.

PR #597, opened 2017-02-25
New "Both" Keymode for Save Point command, use both block name and custom key together.

Treiger:
You can use both the Custom and Block Name Keymodes for your Save Point key, which will produce something like (Block Name)_(Your Custom Key). You can even specify the character that separates the two.

I added this because when I named a Save Point step_1 in two different blocks, the wrong block got executed when I loaded it back. This makes it easier to have unique Save Point keys given that your block names are dissimilar.


chrisgregan:
Nice addition, thank you!

chrisgregan:
btw I made a small change. The new mode is now called "Block Name And Custom" and the summary for the command shows the actual key that will be used instead of "<Block Name>..."

Still works exactly the same way.

PR #601, opened 2017-03-02
Add portrait DimColor

MorningFunGame:
Add portrait DimColor, then user can choose which color they want to dim the portrait, not only 50% grey.
(I am totally newbie to github, please feel free to tell me any suggestions.)


chrisgregan:
Nice addition, thanks!

PR #602, opened 2017-03-19
Write to Fungus

Nilihum:
Created a file allowing for Fungus to write into Playmaker global values. It can currently only write from the inspector and not from flowchart variables. I tried to follow a tutorial[1] on this, but it only shows how to take in the values by hardcoding the variable names, Any help you could give on how to allow the user to write the name of the flowchart variable into the inspector and have Fungus take in that variables value to copy to Playmaker would be greatly appreciated.

[1]https://youtu.be/uRh7INbed2I?t=1219


chrisgregan:
Tried to take a look at the example scene there but the Playmaker FSM appears to be corrupted or something. What happens if you open it up in a clean project on your end?
screen shot 2017-03-22 at 22 45 56

chrisgregan:
To answer your question, looks like you could just use the string property as the name of the variable to lookup in the flowchart, like this.
'''
public void WriteString(string destination, string variableKey)
{
    var stringValue = flowchart.GetStringVariable(variableKey);
    FsmVariables.GlobalVariables.GetFsmString(destination).Value = stringVal;
}
'''

Same for the other types. It is possible to display a variable picker (like in the Set Variable) command but that needs a custom editor class. You can take a look at SetVariable.cs and SetVariableEditor.cs for an example of how that works.

Nilihum:
The Playmaker and Fungus menus are not coming up in the toolbar at the top. That only happens when there's an error in the code, right? Does this happen with my older Pull Request, the one without the "Write to Fungus" script?

chrisgregan:
Could be it alright - I'll try grabbing everything clean again

Nilihum:
I think that the menus didn't load on my end due to simply not having imported them yet, not due to errors.

Nilihum:
I've got no errors after importing Playmaker, but the import button for Fungus seems to be missing from the Asset Store. Any idea what's wrong? Can I DL from github and import that?

chrisgregan:
Yep, go to the releases section on the Fungus GitHub and get the latest
version there.

Nilihum:
I DL'd Fungus from the releases section but the Fungus-Playmaker Integration is missing?

Nilihum:
I've opened the example again in a new project. Everything is fine, except a global variable is missing, but I don't think anything can be done about that. Did you get it working on your end? I think (hope) you might have a problem with Playmaker.

Nilihum:
Regarding the solution you gave me, it worked perfectly well for the string, but not for other variables, where I'm trying to read a string into an Integer, Float, or Bolean, and unable to cast the variable. Any further advice would be greatly appreciated.

chrisgregan:
The Flowchart class has other variable accessor functions like
GetBooleanVariable(), etc. Bool, Int, Float and String versions are
provided, as well as a generic GetVariable method for accessing other
supported variable types. Take a look through Flowchart.cs to see them or
you could check the API docs in the documentation

Nilihum:
Thanks again. Did you find out what was wrong with the example scene?

Nilihum:
Just want to ask here to clear things up. First, how does the flowchart handle concurrency, if it can at all. Second, I've noticed that Fungus has some similarities with decision trees, state machines, and behaviour trees, but is it actually one of them?

chrisgregan:
Sorry, been busy with work for the last couple weeks. Fungus is probably closest to a Behaviour Tree I guess, but it's not formally modelled as one. I tend to think of it as a visual coroutine editor, where each Block is basically a coroutine. Multiple Blocks in a Flowchart can be executing at the same time, but each Block can only have one executing instance at a time. Without this restriction we'd need to create duplicates of each Block for each running instance which would be pretty heavy on performance and also make it harder for users to reason about what's going on.

Nilihum:
Thanks, I had thought it was a bit more like a decision tree, but you  cleared it up. The deadline for the code was Friday, by the way, and for the write up it was today. Presentation on Wednesday. Hope it goes well and thanks for all your help.

chrisgregan:
Nice! Well done on completing the project and good luck with the presentation. Hoping to get a new release of Fungus out soon with the new PlayMaker extensions in - thanks for all the hard work!

PR #607, opened 2017-04-12
Make additional methods in SayDialog virtual

konistehrad:
I often have to extend SayDialog per-game, and in particular often have to do work in Awake. While I was in there, I figured I'd bring the rest of the protected methods into virtual, as I can definitely see use cases for overriding `GetWriter` and `OnEnable`.

The following methods have been made virtual in this PR:
- Awake
- OnDestroy
- GetWriter
- GetCanvasGroup
- GetWriterAudio
- Start
- OnEnable


chrisgregan:
Looks good, thanks!

PR #609, opened 2017-04-25
Lua conditional commands

konistehrad:
Allow Lua to be used in Conditional commands (as `LuaIf` and `LuaElseIf`), and allow Conditional commands of multiple types to see and respond to each other!

This PR comes with at least one nasty compromise: the `Condition#IsElseIf` property is kind of a bummer, but its addition allowed me to preserve the class hierarchy and DRY out the `ElseIf` code quite nicely (also reused in `LuaElseIf`).

Tests are written and pass:
image


chrisgregan:
Works great here, thanks very much for the contribution.

The IsElseIf property seems pretty resonable to me :)

Issue #618, opened 2017-07-05
Prefab Variable Dereferencing 


DLong94:
To reproduce:
1. Create new flowchart into scene
2. Create prefab with flowchart
3. Create a block and add the same command to both prefab and flowchart (involving a variable)
4. Select the prefab and assign the variable
5. Immediately after, select the flowchart object in the scene and assign the variable in the command
6. Save the scene
7. Check the prefab command and variable
8. Variable is dereferenced

Behaviour:
Variable that is set on the prefab is dereferenced and when viewing the prefab command there is no variable selected.

Cause:
This seems to happen when you switch from selecting the prefab to another flowchart object in a scene immediately and saving the scene.

Please find attached a simple game project that exhibits the bug.
[FungusInspectorBug.zip](https://github.com/snozbot/fungus/files/1123229/FungusInspectorBug.zip)

stevehalliwell:
I've done some preliminary digging around in the debugger and it seems that when changing target flowchart the blockinspector is destroyed for the current, then recreated for the current before switching to the next. I'm chasing that down now.

I at first thought that it might have been the commandeditor caching in BlockInspector but forcing it constantly had no effect.

Anyone with more experience with the FungusEditors have any thoughts?

chrisgregan:
Hey Steve, thanks for the bug report and PR. Will look into it this week.

chrisgregan:
Tried out your fix in Unity 2017.1 and it works great - thanks!

PR #620, opened 2017-07-15
Force recreate BlockEditors when target flowchart object changes

stevehalliwell:
Added logic to handle destroying the blockdrawers when the target flowchart changes underneath them. A  nuclear option perhaps but solves Prefab Variable Dereferencing #618


chrisgregan:
Looks like a good fix. I want to do a bit more testing here to check it doesn't break anything else (also doing some testing on 2017.1 atm). Hope to merge it soon.

PR #624, opened 2017-07-26
fixes #612

### LINKED ISSUE #612: Camera->MoveToView and Camera->FadeScreen interference
To reproduce:
1-add new flowchart
2-add command Fade screen (duration 1, alpha 1 wait until finished **unchecked**)
3-add command move to view (set a view, assign, etc)

expected behavior:
camera will fade and move to screen simultaneously

actual behavior:
only move to view will be executed.

Remarks:
-just to make sure disable move to view. Fade screen will be working again.

-after inspection, it is due to `StopAllCoroutines` in `CameraManager.cs` line `428`.
One possible fix is, not to stop all coroutines, only stop the ones that are specific to Panning, not fading
Another fix is to separate these functionalities, so they don't interfere
Finally, another possible fix is to redesign the lerp in a way to allow for mixing.

Something like this:
* remove `StopAllCoroutines` in `CameraManager.cs` line `428`.
* change PanInternal function in a way to add only the delta changes, instead of overriding all values:

          ```
      if (camera != null)
                {
					float deltaOrthographicSize = Mathf.Lerp (startSize, endSize, Mathf.SmoothStep (0f, 1f, t)) - camera.orthographicSize;
					Vector3 deltaPosition = Vector3.Lerp (startPos, endPos, Mathf.SmoothStep (0f, 1f, t)) - camera.transform.position;
					Quaternion deltaRotation = Quaternion.Inverse(camera.transform.rotation) * Quaternion.Lerp(startRot, endRot, Mathf.SmoothStep(0f, 1f, t));
					camera.orthographicSize += deltaOrthographicSize;
					camera.transform.position += deltaPosition;
					camera.transform.rotation *= deltaRotation;

                    SetCameraZ(camera);
                }
```
did not make a pull request yet, just to check if this would break anything elsewhere, as im new to the fungus project.

iBicha:
I think this is the fastest way to fix it, and without affecting the current behaviour, by separating coroutines for panning and fading. 
Having both handles allowing to cancel one effect without affecting the other

### DIFF
--- a/Assets/Fungus/Scripts/Components/CameraManager.cs
+++ b/Assets/Fungus/Scripts/Components/CameraManager.cs
@@ -42,6 +42,10 @@ namespace Fungus
         protected View swipePanViewB;
         protected Vector3 previousMousePos;
         
+		//Coroutine handles for panning and fading commands
+		protected IEnumerator panCoroutine;
+		protected IEnumerator fadeCoroutine;
+
         protected class CameraView
         {
             public Vector3 cameraPos;
@@ -331,8 +335,7 @@ namespace Fungus
                     view.ViewSize);
                 pathList.Add(viewPos);
             }
-
-            StartCoroutine(PanToPathInternal(camera, duration, arriveAction, pathList.ToArray()));
+			StartCoroutine(panCoroutine = PanToPathInternal (camera, duration, arriveAction, pathList.ToArray ()));
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


chrisgregan:
Thanks iBicha, looks like a good fix. Just getting through a work deadline at the minute, will review and merge soon.

chrisgregan:
Thanks for the bug fix!

PR #628, opened 2017-08-20
Added Fungus Commands for most of MathF

stevehalliwell:
Added Math demo scene


chrisgregan:
Another really useful addition - thanks!

PR #629, opened 2017-08-21
Add Missile Command Scene and Commands and Events

DLong94:
This change includes some commands and events which I have written that may come in handy when creating a game with mechanics similar to Missile Command.

**Commands**
- Find GameObject using Tag
- Get the closest GameObject using Tag
- Get current position of GameObject
- Get cursor position of GameObject
- MoveTo command specifying speed
- MoveTo command specifying Vector3 direction
- Generating a random Vector3
- Scale GameObject
- Set Tag
- Spawn Object using Transform
- Vector3 Operations

**Events**
- Collision Event

### DIFF
--- a/Assets/Fungus/Scripts/Commands/Destroy.cs
+++ b/Assets/Fungus/Scripts/Commands/Destroy.cs
@@ -18,6 +18,7 @@ namespace Fungus
     {   
         [Tooltip("Reference to game object to destroy")]
         [SerializeField] protected GameObjectData _targetGameObject;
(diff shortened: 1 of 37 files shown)


chrisgregan:
A similar example game & commands have already been added in for the last release -
see FungusExamples/Playground/Defender by Steve Halliwell.

I think it's better not to have 2 examples in the project that do the same thing, so if it's ok I'm going to close this pr without merging it. Maybe you could put this project into it's own public repo on github and post it on the forum?

PR #643, opened 2017-09-29
Added ability to show variable list in the Flowchart Inspector

stevehalliwell:
-Refactored DrawVariablesGUI
-Added bool to FungusEditorPreferences, that hides the variable list in the flowchart inspector

Disabled by default in the FungusEditorPreferences.
Speeds up workflows that require tweaking a number of Fungus Variables on a number of Flowcharts and inspecting Fungus Variable state if the FlowchartEditor is not visible (low resolution screen, behind game view etc.)

image

### DIFF
--- a/Assets/Fungus/Scripts/Components/Flowchart.cs
+++ b/Assets/Fungus/Scripts/Components/Flowchart.cs
@@ -80,9 +80,6 @@ namespace Fungus
         [Tooltip("The ExecuteLua command adds a global Lua variable with this name bound to the flowchart prior to executing.")]
         [SerializeField] protected string luaBindingName = "flowchart";
 
-        [Tooltip("Draw the variable gui in the inspector.")]
-        [SerializeField] protected bool showVariables = true;
-
         protected static List<Flowchart> cachedFlowcharts = new List<Flowchart>();
 
         protected static bool eventSystemPresent;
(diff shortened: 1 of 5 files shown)


chrisgregan:
Very nice! that's a much better way to manage long lists of variables.

Couple suggestions:
1. Could displaying the variables be controlled solely by the property in the Flowchart inspector? The bool property in the editor prefs seems a bit redundant if you can control it from the Flowchart anyway.
2. Maybe shorten the property name in the inspector to something like 'Show Variables' and an enum picker with two options 'Flowchart Window' & 'Flowchart Inspector'. The default inspector window is pretty narrow so it's usually a good idea to keep property names short to stop them appearing clipped.

chrisgregan:
1. Tried it out here on Mac with Unity 2017.1.1. The layout seems to be a bit off on the variables reorderable list?
screen shot 2017-10-08 at 10 17 43

2. Would it make sense to just always list the variables when a Flowchart is selected? Is there much benefit to being able to toggle the variable display on / off?

3. There should probably be a heading label "Variables" above where the variables list appears in the inspector to explain what the section is about. Right now when you don't have any variables added yet you just see a mysterious + button :)

On the subject of reorderable lists, I'd love to replace the Rotorz ReorderableList control we're using with the built in Reorderable List API. The Rotorz one is really inefficient and generates a load of GC allocs. The Unity API wasn't available when I was writing the Flowchart window and command inspectors unfortunately. Big job though!

stevehalliwell:
1. I tried it on my mac here on 2017.1.0f3 and it doesn't do that, I'll have to update to see if its changed something.

2. Yep, can definitely get rid of the option to hide the variables. I was only offering it so that if users wished, they could keep everything looking the way they are accustomed to.

3. HA, that is an ominous plus sign on its own. I'll see if there is an option to get the reorderable list to show it's header even when empty.

chrisgregan:
Just checked the latest version of this an it's working fine on my Macbook Pro now. Good to commit I think!

Issue #572, opened 2016-11-21
Wait for Voiceover option with Say dialog


andrewtraviss:
I'm using Say commands for a voiceover-only scene, which is causing problems because Say can only wait for input or wait for the writer to finish writing out the text. In this case, I need Say to wait for the voiceover to complete and then proceed.

There are two workarounds I'm using:
1. Use Play Sound - but this causes some other issues with audio fading, and I was hoping to be able to use the text portion of Say as an optional captioning system
2. Insert Wait commands after each Say and use {s=0} for all say blocks, which just adds a ton of script bloat and busywork, especially for a longer script.

PR #645, opened 2017-10-05
Added TextTag WVO WaitForVoiceOver #572

stevehalliwell:
Say has a Wait For VO flag to add wvo text tag like it does the wi tag
Added SayHolds demo scene to show how the combinations of waits interact


chrisgregan:
Tried the demo scene out here and reviewed the changes, looks great!

PR #652, opened 2017-12-02
Added GlobalVariables

stevehalliwell:
-'Global' variable scope
-Access is routed at run-time through the GlobalVariables manager attached to the FungusManager
--If it does not already exist it is created there and given the starting value from the local creator

Allows for communication and sharing of data between flowcharts and scenes without having to route through save and load variable. Can also be used as a method of 'send' variables along with flowcharts messages.

To use, simply create the variable as per normal, set it's scope to Global. Every flowchart that has a variable of the same name and scoped as global is now using the same underlying variable in the global variables.

The GlobalVar demo, shows multiple flowhcharts interacting with the same int global variable.


FatPuzo:
Hello, Steven!
I've ran to some issue, after updating fungus files.
What I did: 

- Downloaded and unpacked code from github.
- Deleted original Fungus and FungusExamples folders and replaced with one's from github.
- Started my test project

After click on <New Block> in flowchart I get some errors in hierarchy. Each time I click <New Block> more FungusManager objects get spawned.
Also I get uneditable block commands in Inspector (and that block is not complete)

hasglob

I'll also post what it looks like with original files:
noglob


P.S.: Steven, the idea of global variables is perfect, thank you for the contribution!

PR #654, opened 2017-12-17
Added GenerateFungusVariable tool

stevehalliwell:
-Auto generates much of the boilerplate code required to add a new variable type to fungus
-Auto generates a Property Command to get and set the understood fields and properties of that type
Added ability to set a fungus variable as preview only in the Variable List
Re-gen Command Reference
Added Quaterion and Matrix4x4 fungus variable types using the new tool

image


chrisgregan:
Cool! 

I tried this out by generating a variable and a property for a Unity Vector4. Looks like it worked ok, though the property code seems to indent a bit wierd:
screen shot 2018-01-03 at 23 27 27

Also the Vector4 values don't seem to display correctly in the variable list.
screen shot 2018-01-03 at 23 28 39

stevehalliwell:
My thinking with the tool is that it gets you most of the way there. 
I was being lazy and using the Visual Studio auto format on the scripts after generating. I can make the script better at correct formatting as part of the generation though.

The test cases I used were Quaternion and Matrix4x4 both of which I manual set to preview only in the variables window as manually editing them would not be advised. I'll see if I can find a clean way of dealing with that in the tool and beyond that of dealing with variables that want to do foldouts.

stevehalliwell:
The foldout for the Vector4 appears to be an ongoing issue for the Reorderable list. We can get it to draw over the top of others by turning on include children in all the EditorGUI PropertyField calls. But identified [here](https://bitbucket.org/rotorz/reorderable-list-editor-field-for-unity/issues/28/broken-gui-on-struct-ordering) Unity is not correctly transmitting the state of isExpanded for serialisedProperty array elements.

stevehalliwell:
I've found some odd behaviour I need to track down and a few other features I think it should have, so closing this, at least for now.

PR #656, opened 2017-12-21
Upgraded project to support Unity 2017.3.0

charblar:
Didn't include API upgrades which changes captilization to UnityEditor Events (EventType.mouseUp to EventType.MouseUp)


chrisgregan:
Thanks charblar. I made a bit of a mess of integrating this PR in git, but the changes are in now.

PR #659, opened 2018-02-05
Correcting the repeat play of music

KVinS:
Corrected:

We will not hear the music if we try to restart, the stopped track.
- Play (A)
- StopMusic
- Play (A)
No result. (Line 30 - audioSource.clip == musicClip)

### DIFF
--- a/Assets/Fungus/Scripts/Components/Flowchart.cs
+++ b/Assets/Fungus/Scripts/Components/Flowchart.cs
@@ -87,6 +87,14 @@ namespace Fungus
         protected StringSubstituter stringSubstituer;
 
         #if UNITY_5_4_OR_NEWER
+        protected virtual void Awake()
+        {
+            CheckEventSystem();
+
+            UnityEngine.SceneManagement.SceneManager.activeSceneChanged += (A, B) => {
+                LevelWasLoaded();
+            };
+        }
         #else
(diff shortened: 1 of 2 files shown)


KVinS:
Adding methods to work with blocks, as with variables:
- Check existence
- Execute the block, even if it does not exist.
Without generating exceptions.

Analogues:
- HasVariable
- SetVariable

stevehalliwell:
Thanks KvinS, the first commit here is a good fix. I'm not sure what the second commit is doing? Is it related to repeat play music?

KVinS:
In the second commit two methods are added.
- HasBlock
- ExecuteIfExites
They are needed for work when we are not sure of the existence of a block.
Now, such methods are only for variables.

chrisgregan:
The changes here are good, but there were some problems with the code so I reimplemented and committed the same changes. The audio property on Monobehaviour is deprecated, so I changed MusicManager to cache the AudioSource in a protected variable and set the clip to null on that instead.

The changes to Flowchart.cs look quite useful, but the PR changed all the line endings in the file so it appeared in the diff like all the lines had been deleted and added again. I copied and pasted the new methods locally and unified the line endings so the correct edit history is preserved.

btw it's generally better to submit one PR for each feature / bug fix as it makes it easier to review and merge them individually.

Thanks for the contribution!

Issue #657, opened 2018-01-14
Save system not working in VN with multiple Flowcharts


jhm-ciberman:
Hello. I have a problem. I am developing a linear visual novel. The visual novel is complete. I have added the Save System. When I use the “Load” button in the GUI or when i check the Autoload button in the SaveMenu component and hit play, the game loads the scene indefinitely and never starts. The Save game example provided with fungus works fine. So, I think that the problem is related to:

1 – I am doing something wrong and the error it's not handled by fungus

2 – The Save System does not works for larger projects.

Here is the Unity source code of the project:

https://github.com/Pilii62/vn-arritmia

Feel free to clone the repo and examine it.

Thanks.

if you need something more, please tell me.

PS: The other bug that I cannot solve, it's that when I press any SaveMenu button, the click input goes both to the SaveMenu button and the SayDialog (Dialog Input). If you can help me with that I will be glad. Thanks.

Formum link to the issue: https://fungusgames.com/forum#!/support-questions:save-system-not-working-in

MeMark2:
The problem seems to come from the fact that you nested the Save Menu in another game object (your GUI game object).

The Save Menu is a singleton that relies on the [DontDestroyOnLoad](https://docs.unity3d.com/ScriptReference/Object.DontDestroyOnLoad.html) functionality. Unfortunately though, [DontDestroyOnLoad doesn't work if the script that uses it is inside of a parent that doesn't call the method](https://answers.unity.com/questions/452864/dontdestroyonload-child-object.html).

In other words, making the Save Menu a child of another Game Object breaks it.

I think this is a problem with the design of the Save Menu script. My instinct tells me that the singleton should rely on static class members more instead of using the DontDestroyOnLoad functionality and treat the instantiated singleton more like a link between the usual Unity work flow and the static aspects of the class (I'm basing this thought on [this stack overflow question](https://stackoverflow.com/a/519530)). However, I don't have a whole lot of past experience working with singletons, so I'm not sure if that would really be the best approach.

Alternatively, the Save Menu could un-parent itself at some point during runtime in order to prevent it from being destroyed along with a parent, but I don't like this approach since **a)** it would be confusing for most users to see the Save Menu change its own hierarchy while the game is running, and **b)** it prevents the user from nesting the Save Menu within other objects or panels if they wish to customize where the Save Menu shows up.

chrisgregan:
I've merged the pull request that fixes this issue

PR #661, opened 2018-03-11
Account for SaveMenu DontDestroyOnLoad problems

MeMark2:
Threw in some infinite loop prevention when the SaveMenu is nested inside other GameObjects (and therefore the DontDestroyOnLoad method doesn't work). Also logs a more context-aware error when the SaveMenu is nested inside another GameObject.

This is my first time contributing. Please let me know if there's anything I could do better.

For further details, see [this issue](https://github.com/snozbot/fungus/issues/657#issuecomment-372126664).


chrisgregan:
Nice fix, thanks!

PR #663, opened 2018-03-16
Blocks.md Documentation Correction

Rubikhan:
Number 6 of the section "Connections between Blocks" in blocks.md incorrectly referred to "Block2" instead of "Start". Changed to reflect what is actually displayed in 3_menu_call.png


chrisgregan:
Thanks for the fix!

PR #671, opened 2018-04-09
Added Fungus Priority Signals

stevehalliwell:
Intended to allow fungus to notify the outside world that it is doing something of priority so the outside world can pause
-Tests
-Commands

### DIFF
--- a/Assets/Fungus/Scripts/Commands/Priority.meta
+++ b/Assets/Fungus/Scripts/Commands/Priority.meta
@@ -1,0 +1,10 @@
+fileFormatVersion: 2
+guid: 61d2004f0aa3f1847beb167296c897d8
+folderAsset: yes
+timeCreated: 1523180219
+licenseType: Free
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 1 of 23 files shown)


chrisgregan:
Looks like useful functionality, some suggestions about the FungusPriorityCurrentCount command. 
1. Class name has typos and should exactly match the name in the command info attribute. 
2. Maybe rename the command to something more descriptive like GetPriorityCount?
3. The command description doesn't match what the command does.
4. These command are using the default white color, maybe assign a pastel shade to them?
5. Use a VariableProperty attribute to display the variable picker correctly, e.g.
```
        [VariableProperty(typeof(IntegerVariable))]
        [SerializeField] protected Variable outVar;

        public override void OnEnter()
        {
            IntegerVariable variable = outVar as IntegerVariable;
            variable.Value = FungusPrioritySignals.CurrentPriorityDepth;

            Continue();
        }
```

I'd suggest grouping these commands under a single high level 'Priority Signals' group in the command list to help keep things organised. 

On a side note related to that, some of the high level groups only have one command under them. What about creating a Misc high level group and group these together neatly under it? e.g. Misc > Instantiate GameObject, Misc > Get Input Axis, Misc > Reload Scene, Misc > Set Transform Property - that kinda thing?

PR #672, opened 2018-04-14
Force the hierarchy icon list to refresh when a script a is loaded/ch…

stevehalliwell:
…anged, fixes delay when icons disappear after a code change is made by no change is made in Unity itself.

### DIFF
--- a/Assets/Fungus/Scripts/Editor/HierarchyIcons.cs
+++ b/Assets/Fungus/Scripts/Editor/HierarchyIcons.cs
@@ -25,8 +25,11 @@ namespace Fungus
         //sorted list of the GO instance IDs that have flowcharts on them
         static List<int> flowchartIDs = new List<int>();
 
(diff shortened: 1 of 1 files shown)


chrisgregan:
I'm not familiar with this issue - how can I replicate getting the icons to dissappear?

stevehalliwell:
Found in 2017.2 & 2017.3 unsure if it persists beyond those.

With a scene open in Unity that has Flowcharts in it.
Add a new c# script to project.
Open that script in VS or similar.
Make a new script or a script change, anything that causes a recompile. 
Switch back to Unity.
Touch nothing.
Unity will recompile and the fungus hierarchy icons will disappear.

Suspected cause;
Unity is correctly recreating the static Hierarchy Icons instance and notifying it do draw the elements but it does not resend the changed message on script changes.

Solution;
Keep track of the reload scripts ourselves and when detected, invalidate our fungus object cache.

PR #673, opened 2018-04-15
Conversation options

stevehalliwell:
Follow up from [the forum](http://fungusgames.com/forum#!/general:extending-the-conversation)
Allows a Conversation to control a similar set of options as a Say command.
-On the conversation command itself you can set defaults for fade, wait for input, wait delay, extend previous
-Each item in the conversation can now also set [clear | noclear] [wait | nowait] [fade | nofade] before the story text


chrisgregan:
This works really well, nice one. How about extending the Conversation example scenes to show using some of the new tags?

PR #694, opened 2018-06-28
Adding Random Sound Command and Ambiance Sounds

HectorCastelli:
This update will add:

1. A new command that will allow users to play one randomly selected audio clip from a list of audio clips.
2. A new command that allows users to queue up ambiance sounds, that can be looped or not and will be played over the music. Allowing for ambiances to be queued up over the existing music without stopping it.


HectorCastelli:
Hello Chris, I will do that in the next two to three days. Should I ping you when it's done?

chrisgregan:
Yes, please do.

HectorCastelli:
@chrisgregan I got it! I just updated the text. Sorry for the delay!

PR #695, opened 2018-06-29
Variable text selection

stevehalliwell:
Writer and Menu now use TextVariationHandler which allows for [ink style text selection](https://github.com/inkle/ink/blob/master/Documentation/WritingWithInk.md#6-variable-text).

Sections are marked with []
sub elements divided by |

These are sequences by default but can be made to loop, once only or random each time with control characters. They can also be nested within each other.


chrisgregan:
Very cool feature!

chrisgregan:
One suggestion, how about updating the Conversation topic in the docs to explain the new tags?

PR #698, opened 2018-07-22
Store the profile name list in player prefs

awiebe:
This should store the list of profiles as a base64 string, but needs further testing.
Call deserializeProfileList() if you want to extract the list of profiles(e.g. for some kind of load/save profile command)

I'd put it on the develop branch if one existed.

### DIFF
--- a/Assets/Fungus/Scripts/Commands/SaveVariable.cs
+++ b/Assets/Fungus/Scripts/Commands/SaveVariable.cs
@@ -2,6 +2,9 @@
 // It is released for free under the MIT open source license (https://github.com/snozbot/fungus/blob/master/LICENSE)
 
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


chrisgregan:
Hi aweibe, thanks for the PR. Could you explain the benefit of storing the profile name list as a base64 string? Is it possible this change might cause problems for users with existing save data upgrading from an previous version of Fungus? There also appear to be some compile errors in SaveVariable.cs currently.

Cheers,
Chris

awiebe:
@chrisgregan
Fungus has a way to set the active profile using a key, but doesn't have a way to retrieve the names of these profile keys, at least as far as I could see.

So if for example you wanted to present the user with a list of save-games, you would need to grab the list from the `PlayerPrefs`

PR #703, opened 2018-08-03
Fixed a bug when click KeypadEnter on BlockEditor

qjf1992:


qjf1992:
The operate will throw an exception.

qjf1992:
This is my first pull request and some question here.
I never change the last line of file, but the diff happened. 
I had compared when commit to my fork. No diff at the last line.
Is there any wrong opt I do?

Thank u very much If you can help me... TAT

stevehalliwell:
Thanks. Good fix.

The last line change is most likely due to your text editor / IDE / git client enforcing a newline at the end of the file

PR #704, opened 2018-08-04
Unity reorderable

stevehalliwell:
Remove Rotorz and use Unity's InternalReorderable list in its place.


chrisgregan:
Awesome job Steve! One small thing, the command name and info text seem to be slightly misaligned vertically? Seem to be off by a pixel or two on some rows.

screen shot 2018-08-13 at 23 01 45

PR #705, opened 2018-08-07
First person demo

stevehalliwell:
Simple interactions between Fungus and Unity's Standard Assets for FirstPersonController and Prototyping.
Makes use of Block Reference and Fungus Priority Signals.


chrisgregan:
Cool demo! Including the full First Person Character standard asset as part of Fungus might be a problem, I'm not sure [that's allowed](https://support.unity3d.com/hc/en-us/articles/206336915-Can-I-submit-an-asset-that-uses-Unity-Standard-Assets-). If we can do it, it should probably live in the Fungus/Thirdparty folder along with the other third party code.

If we can't ship the FPC with Fungus then perhaps this demo could be bundled up into a .unitypackage and we link to it from a blog post & the docs?

stevehalliwell:
I'll remove all of the Unity Standard Asset elements and recreate it. Not a lot to do was just faster to use existing Unity stuff.

chrisgregan:
Looks good, thanks!

PR #706, opened 2018-08-29
Doco for Text Variation and Variable overview

stevehalliwell:


chrisgregan:
Nice improvements, thanks!

PR #708, opened 2018-09-06
Specifically identify choice blocks as hexagons 

mattboch:
Currently the documentation refers to blocks as either rectangles or polygons. This is a strange choice as rectangles are subset of polygons; a rectangle is a type of polygon. Thus, identifying the choice blocks as polygons is unnecessarily confusing, especially for the color blind. Given that the choice blocks are six-sided, they can be more specifically identified as hexagons.


chrisgregan:
Thanks for the improvement!

PR #709, opened 2018-09-14
Command searchable pop up

stevehalliwell:
Adds Option to Fungus Editor Preferences to enable searchable, partial match popup windows in place of the existing Add Variable, Select Event and Add Command Generic Menu versions. Option is disabled by default.

Search box shows partial matches in list below, which is scroll-able and shows command and event descriptions as tooltips on hover in their respective menus.

var menu

event menu w partial

command w partial w tooltip visible


chrisgregan:
Yet another cool addition :)

Any reason why we wouldn't just switch to this UI instead of making it an experimental feature? Looks like it's a better workflow than the old way, but still similar enough that it shouldn't be confusing for new users.

I did find a couple issues when I tried it out though.
1. Selecting a command in the drop down list adds a different command than the one selected.
2. Shrink the inspector to its minimum width. The add command controls are pushed offscreen.
screen shot 2018-09-17 at 21 38 03

stevehalliwell:
I'll chase down those issues this weekend.

I would eventually like to see it replace the generic menus as the default, as I believe it will increase speed of use and reduce total clicks. In particular the hover tooltips will be a great help to new users. Will probably still want a way to toggle parts of it off, as there will be Fungus users with years of muscle memory for the GenericMenu based layout.

Putting it behind the experimental toggle comes from a couple of places; 
- allow users to try the change in workflow and provide feedback before it is forced upon them
- get bugs related to specific window layouts or other tools they use that conflict with these changes as they are often tough to find or predict
- use experimental in to indicate that the feature may change as bugs and feedback come in.

chrisgregan:
Finally resurfacing after a crazy couple of months in work. I think this feature works really well, maybe we could enable it by default and have the option in preferences for anyone that wants the old behaviour?

PR #718, opened 2018-12-04
TextMeshPro support for Writer and MenuDialog

stevehalliwell:
Introduces a TextAdapter class to bridge between a Fungus element that wishes to control or update a text element on screen and the actual text component type.
This TextAdapter is then used by Writer and MenuDialog allowing for both to have any mix of text element supported by the TextAdapter. 

Internally the TextAdapter is the previous Writer specific code refactored out. As that already attempts to find a Component with a 'text' property, we can use TextMeshPro


chrisgregan:
Hey Steve, looks like this works well. I was thinking though, TextMesh Pro is now installed as a default package - do we need to include it in the Assets folder as well?

stevehalliwell:
Had a quick check. The TMPro Unity package appears to only give partial functionality. I assume it's to guarantee to parts of Unity that the code will exist to compile against but it doesn't appear to have everything required to use TMPro in a scene. Once you use any of TMPro this popup appears when ever it is accessed.

image

The folder in Fungus is the result of Import TMP Essentials. 

If a user didn't want TMPro they could delete the TMPro scene and the TMpro folder at the root.

chrisgregan:
Ok looks good, I've merged this pull request. I'm going to move the TextMesh Pro folder under Fungus/Thirdparty just to keep the folder structure neat.

PR #725, opened 2018-12-28
New Call mode, StopThenCall

stevehalliwell:
Relates to [this forum thread](http://fungusgames.com/forum#!/support-questions:bug-with-call-command-if-ca)

Allows for circumstances where you want to call a block that completes immediately multiple times such as in recursion.
Removes the need to put Wait 1 Frame commands in a number of cases
More verbose logging in cases where a block cannot be called as it is already running, which seemed to cause the initial confusion in the linked thread.

### DIFF
--- a/Assets/Fungus/Scripts/Commands/Call.cs
+++ b/Assets/Fungus/Scripts/Commands/Call.cs
@@ -18,7 +18,9 @@ namespace Fungus
         /// <summary> Continue executing the current block after calling  </summary>
         Continue,
         /// <summary> Wait until the called block finishes executing, then continue executing current block. </summary>
-        WaitUntilFinished
+        WaitUntilFinished,
+        /// <summary> Stop executing the current block before attempting to call. This allows for circular calls within the same frame </summary>
+        StopThenCall
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


chrisgregan:
To test this out, I opened the Enter Name example, selected the End Edit block and changed the Call command to use Stop the Call mode. The command executes ok, but it seems to trigger the 'ReadText cannot be called/executed, it is already running.' warning, which doesn't sound correct?

stevehalliwell:
I should have been clearer, it doesn't Stop the called block from running, it stops the caller from 'running' before making call. 

In cases where we have block A and B, where A is going to call B,

The existing Stop method puts the Caller into a state where the next time it is ticked it will move into the idle, not running, state. This behaves like we would expect from a call stack, kinda. We get something akin to;
```
A Calls B
  B Runs all non waiting commands
A StopsParentBlock, putting its index to max int

//next tick
A Completes, returns to idle
```

The new StopThenCall method finalises and puts the Caller into the idle, not running, state before B is called. This behaves vaguely like a tail call optimisation (a bit of a stretch, I know). We get something akin to;
```
A StopsParentBlock and Completes, returns to idle
A Calls B
  B Runs all non waiting commands
```

In the case of Enter Name, if you click the OK button (rather than hitting enter), both the block with the Button hooked up as its event and the block with the End Edit hooked up fire. A block cannot be executed while it is already running, previously this was silently eaten, now there is a log to help assist in cases where the call is desired and known to be hit but not behaving as desired.

chrisgregan:
Ok, looks good

PR #731, opened 2019-01-10
Set background image for Menu-Item

ducksonthewater:
-- sorry, I am fairly new to GitHub, and was not able to create pull-requests the right way --

I added a simple sprite-reference to set the background image of a newly created menu-option. An example-target is given in the new prefab, the command "Narrative -> Menu" was extended.
This way, users can create text-only (as of now), image-only or text/image-menus by editing the menu-prefab.

-- please ignore the scene Management (092841c), I added and changed some things, and will create a new pull request --

The proposed Scene-Commands can be used for loading sublevels in the background, e.g. when entering a building:
- before opening the door in game, call SceneLoadAdditive (check "Show Scene When Loaded")
- after having closed the door, call SceneRemoveAdditive

SceneLoadAdditive - loads a scene additively by name. Optional: display scene after loading, call block in Flowchart. If you do not call SceneActivateAdditive, Unity may crash, as the async-operation will be broken. 
SceneActivateAdditive -  activate a scene by its name. Useful, if you want to load things, and activate them later.
SceneRemoveAdditive - removes a scene by its name.

### DIFF
--- a/Assets/Fungus/Resources/Prefabs/MenuDialog_Image.prefab
+++ b/Assets/Fungus/Resources/Prefabs/MenuDialog_Image.prefab
@@ -1,0 +1,1782 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!1001 &100100000
+Prefab:
+  m_ObjectHideFlags: 1
+  serializedVersion: 2
+  m_Modification:
+    m_TransformParent: {fileID: 0}
+    m_Modifications: []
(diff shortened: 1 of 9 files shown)


chrisgregan:
Hey, thanks for sending in the new features! Would it be possible to resubmit these using a separate pull request per feature? Just makes things easier on my end if they're split up.

ducksonthewater:
Ja, sorry for this. I can try. Never did (several) pull requests before.

PR #733, opened 2019-01-15
Command has reference and summary updates

stevehalliwell:
Many commands now have more thorough HasReference implementations.
- Added to some commands, such as iTween commands
- Checking more variables and data in some commands, such as SetVariable
- Added new referenced variable cache list to commands for non trivial checks.
-- This is now used in all commands that used the SubstituteVariable, such as Conversation, on a string variable to show the variables that will be used in the inspector by that substitution.


chrisgregan:
Looks good, thanks!

PR #739, opened 2019-04-04
Added Vector2Int & Vector3int & get set animator

LeieSistal:
Now you can create and play with Vector2Int & Vector3Int
You can also access et set Vector2Int, Vector3Int and Animator via C# Scripts


LeieSistal:
Sorry there is something wrong with my commit

PR #742, opened 2019-05-10
New commands for additive Scenes in Unity.

ducksonthewater:
Edit: All scenes should be self-contained. You can't call blocks in other Flowcharts or access their variables. Two separately loaded scenes cannot contain hard-links to each other. If you need to call other Flowcharts, use messages to all Flowcharts.

Scenes MUST be added to Build Settings.

- SceneActivateAdditive
Activates a loaded scene. Just needs the scene name.

- SceneLoadAdditive
Loads a scene. Optionally activates it, and calls a Fungus block after activation.

- SceneRemoveAddtitive
Remove the added scene. This has to be called, when a Scene is not needed anymore.

- SceneEditor
Helper to see the blocks in a scene, when using SceneLoadAdditive


Assets/Fungus/Scripts/Commands/SayLocalized.cs
```diff
@@ -0,0 +1,210 @@
...
+                    displayText = displayText.Replace(ct.TagEndSymbol, ct.ReplaceTagEndWith);
+                }
+            }
+
+            string subbedText = flowchart.SubstituteVariables(displayText);
+
+            AudioClip audioTemp = voiceOverClip;
+            // this should come from the localization-object itself
+
+            voiceOverClip = Resources.Load<AudioClip>("Speech/" + GameObject.Find("GameSafe").GetComponent<Flowchart>().GetStringVariable("Language") + "/" + audioTemp.name);
```
chrisgregan: This depends on the user having a Flowchart with a specific name and a specific named variable in that Flowchart. It would be cleaner to use the existing SetLanguage command and query a Localization object for the active language. See the localization example for reference.

Assets/Fungus/Scripts/Commands/SayLocalized.cs
```diff
@@ -0,0 +1,210 @@
...
+namespace Fungus
+{
+    /// <summary>
+    /// Writes text in a dialog box.
+    /// </summary>
+    [CommandInfo("Narrative",
+                 "Say Localized",
+                 "Writes text in a dialog box.")]
+    [AddComponentMenu("")]
+    public class SayLocalized : Command, ILocalizable
```
chrisgregan: This command is a modified copy of the Say command with a few modifications. Given the large amount of overlap between the two commands, it would be better to inherit from the Say command and override the parts that are different. This would avoid the need to keep 2 versions of the Say command in sync as bug fixes and improvements are made.

chrisgregan:
Thanks for the PR! The scene loading and Unity Ads support changes look great. I've made a couple comments about the SayLocalized commands which would need to be addressed to accept the PR.

btw it's better to submit each feature as it's own PR so I can check and merge them in individually.

Cheers!

Assets/Fungus/Scripts/Commands/SayLocalized.cs
ducksonthewater: Hi Chris, I reverted my Say-commands within GitHub.
Will have a look on how to improve it.

ducksonthewater:
Great to hear :-)
About the "say", I will have a look.

PR #761, opened 2019-09-18
Added a custom scripted importer for lua files

Taking1n1:
This change uses ScriptedImporter to recognise .lua files and create text assets out of them.

The result is: you can drag .lua files into text asset regions (like in the Execute Lua command) and you can have syntax highlighting with your favourite editor when editing said .lua files (I use sublime, and it's kind of annoying to keep setting the styling to Lua when editing .txt files).

I don't think this is something that can be "tested", so I put an animated gif for proof:

The lua file:
fungus1

Dragging it into text asset slots:
fungus

This is my first pull request onto an open-source project; go easy on me! 😅


stevehalliwell:
Good idea and thanks for the PR.

Would it be possible to achieve this without using the Experimental feature? Just so this doesn't cause compatibility issues when the API changes or the feature moves out of Experimental?

We should also wrap the class in the Fungus.Editor namespace to avoid potential clashes.

Taking1n1:
I haven't pushed it but I've tried a different approach to avoid the experimental ScriptedImporter.

Basically, Lua Scripts accept any object, not just TextAssets 😅 That way they can accept .lua files and then read the text itself.

However there are some problems with this; there's far less validation, it's hacky and there will be significantly more overhead in the lua script loader that loads from Resources/Lua (as it would have to check every file).

Do you think it would be better to push this idea or to wait until ScriptedImporter becomes stable?

stevehalliwell:
Let's use this method then and we'll mark it as Experimental in the change log.

Assets/Fungus/Scripts/Editor/LuaScriptedImporter.cs
```diff
@@ -0,0 +1,14 @@
...
+[ScriptedImporter(1, "lua")]
+public class LuaScriptedImporter : ScriptedImporter
+{
+    public override void OnImportAsset(AssetImportContext ctx)
+    {
+        TextAsset lua = new TextAsset(File.ReadAllText(ctx.assetPath));
+        ctx.AddObjectToAsset("main", lua);
+        ctx.SetMainObject(lua);
+    }
+}
```
stevehalliwell: Insert 
`namespace Fungus.EditorUtils
{
...
}`

Issue #760, opened 2019-09-18
Fix for BlockInspector Bug for Unity 2018+


Michaelwolf95:
In some versions of Unity 2018 or newer, Fugus's BlockInspector would fail to populate the inspector, and constantly throw null references. I know I'm not the only one running into this, and I just found a fix for it. 

Normally I'd just fork, fix it, and submit a pull request, but I don't have the time right now, and I want to make sure this gets out there for people to see. I'll try to come back and submit a fix properly later.

**SOLUTION:**
In `BlockInspector.cs`, simply change the line that instantiates the `activeBlockEditor` to include the `BlockEditor` type as a parameter.

For the latest asset store release, change line 96 of `BlockInspector.cs` from
`activeBlockEditor = Editor.CreateEditor(block) as BlockEditor;`
to
`activeBlockEditor = Editor.CreateEditor(block, typeof(BlockEditor)) as BlockEditor;`

PR #762, opened 2019-09-19
Fix for BlockInspector Crash

Michaelwolf95:
On some newer versions of Unity, the BlockInspector will crash whenever the user selects a block in the flowchart window.
The inspector window will not populate, and the console will throw null reference errors every editor frame as long as the block is selected.
This is due to `activeBlockEditor` never getting instantiated properly. I suspect that this is due to recent API changes that made `Editor.CreateEditor(block)` create a generic Editor object, instead of a BlockEditor. Since the object is not actually the extended type, the reference becomes null after being cast using `as`.

The fix is to change this line to `Editor.CreateEditor(block, typeof(BlockEditor))`

This overload parameter ensures that the object is instantiated using the correct type. This fix should work for all versions, as it appears to have been an API option since at least Unity 5.2.

Tested on Unity 2018.0.1f1 and 2019.2.1f1

### DIFF
--- a/Assets/Fungus/Scripts/Editor/BlockInspector.cs
+++ b/Assets/Fungus/Scripts/Editor/BlockInspector.cs
@@ -97,7 +97,7 @@ namespace Fungus.EditorUtils
                 !block.Equals(activeBlockEditor.target))
             {
                 DestroyImmediate(activeBlockEditor);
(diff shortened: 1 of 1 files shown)


stevehalliwell:
Thanks for the PR.

stevehalliwell:
Fixes #760

Issue #774, opened 2019-11-11
Search should search commands, not just block names


AcademyOfF:
https://imgur.com/iEYTzda

In this image you can see I'm searching for a term that is in a command (a variable in an IF statement). But, the block is still grayed out as if that search term doesn't match.

Does the search only find block names? Is there a way to have it search content inside blocks?

If this feature doesn't exist, what's a good work around? My game is going to have 1000s of blocks, I can't be clicking on them one by one to find where a particular variable was set.

PR #775, opened 2019-11-11
Search should search commands, not just block names

AcademyOfF:
Sorry, I don't know how to use github PRs well.  This is for issue  #774

https://imgur.com/iEYTzda

In this image you can see I'm searching for a term that is in a command (a variable in an IF statement). But, the block is still grayed out as if that search term doesn't match.

Does the search only find block names? Is there a way to have it search content inside blocks?

If this feature doesn't exist, what's a good work around? My game is going to have 1000s of blocks, I can't be clicking on them one by one to find where a particular variable was set.


stevehalliwell:
Thanks for the PR, a good addition.

PR #779, opened 2019-11-20
Improve Performace for Singleton

AndyHan1001:
Use "double checked locking" algorithm to implement singleton for "FungusManger" class, which can improve performace.

### DIFF
--- a/Assets/Fungus/EditorResources/FungusEditorResources.asset
+++ b/Assets/Fungus/EditorResources/FungusEditorResources.asset
@@ -14,56 +14,56 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   updateOnReloadScripts: 0
   add:
-    free: {fileID: 2800000, guid: 24a867d6b4cdda04cb3aa3350f9ec7d0, type: 3}
-    pro: {fileID: 2800000, guid: 4e43c476b4a7a49a08c37378fb01ce99, type: 3}
+    free: {fileID: 0}
+    pro: {fileID: 0}
   add_small:
-    free: {fileID: 2800000, guid: 288aff1a0e5c14fd3a0039d24149df73, type: 3}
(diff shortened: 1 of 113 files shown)


stevehalliwell:
Hey AndyHan, thanks for putting in a PR. Did you have a use case for multithreaded access to the FungusManager?

AndyHan1001:
> Hey AndyHan, thanks for putting in a PR. Did you have a use case for multithreaded access to the FungusManager?

Hi stevehalliwell, I am sorry, I do not have any use cases for that, but "double checked locking" algorithm is a proven method to handle multithreaded cases and guarantee the performance at the same time.

stevehalliwell:
Thanks Andy, I was wondering if this was in response to a specific need or situation you found as my concern is this makes it look like Fungus is thread safe and/or thread aware, when it isn't. 

Related, much of Unity itself will throw errors back at you if you try to interact with it from anything other than Unity's MainThread. If we called FungusManager.Instance on another thread when it was null, new GameObject, DontDestroy, and AddComponent might just throw back errors from Unity.

AndyHan1001:
> Thanks Andy, I was wondering if this was in response to a specific need or situation you found as my concern is this makes it look like Fungus is thread safe and/or thread aware, when it isn't.
> 
> Related, much of Unity itself will throw errors back at you if you try to interact with it from anything other than Unity's MainThread. If we called FungusManager.Instance on another thread when it was null, new GameObject, DontDestroy, and AddComponent might just throw back errors from Unity.

Right, but my pull request is just for performance improvement. No extra function is added. Your original code will always lock the thread if you access instance of FungusManager, which is time consuming. My pull request code only need to lock the thread once, which can improve performance significantly if the instance of FungusManager is accessed frequently.

stevehalliwell:
Yep, I missed that line change in the diff preview. Merging in changes to FungusManager.cs now and trying to avoid all the meta changes.

Thanks.

PR #782, opened 2019-12-02
Made sure my save system works with the latest Fungus

CG-Tespy:

PR #784, opened 2019-12-14
Narrative Log System Replacement for Fungus

CG-Tespy:
Original repo here: https://github.com/CG-Tespy/Fungus-Slot-based-Narrative-System


stevehalliwell:
Hey, I assume same deal as with #782 ?

I should have enough time to go other them during the holiday break.

CG-Tespy:
Yup, same deal. I even have integrations between my Save and Narrative systems in the repo of the former, in the SaveSystemExtras folder :)

PR #785, opened 2019-12-19
Remove previous doco, now lives in the github wiki

stevehalliwell:
Correct Fungus Priority Category to remove space
Added Convert To Github wiki format function to Tools menu
Export of reference docs moved out of BlockEditor.cs


kurtmel:
pardon my complete and total ignorance here, but how do I obtain a complete offline version of the latest Fungus Manual. I am very much an RTFMF kind of person, and an Instructional Designer, NOT a coder, (Although learning, as well as 3D modeling, etc.), so some gentle prodding in the right direction would be greatly appreciated.

PR #791, opened 2020-02-10
Draggable events array

### LINKED ISSUE #238: Drag and Drop event handler to support array of target objects
From the forums: "Would it be possible to add an array to the event targets involved in dragging and dropping? It would be nice, if I have 5 draggable objects, they all halt swipinng when being dragged, instead of having to create 5 different on drag start events."

"Maybe we could do something with GameObject variables. Like have the Event Handler take an array of game objects to work on, and a GameObject variable that will get populated when the event happens so that the commands in the block can act on it. I can see how it would be useful for complex drag and drop setups."

SarveshBajaj:
Fixed #238


SarveshBajaj:
**InvalidOperationException: EnsureRunningOnMainThread can only be called from the main thread**

I am getting the above error when we check whether the target reference is null during runtime in the function ISerializationCallbackReceiver.OnAfterDeserialize, it doesn't seem to affect any functionality, so should I force it to execute only in editor mode?

stevehalliwell:
Can definitely put it to editor only but that won't fix the problem, and because it's throwing it won't actually complete the upgrade from old fields to new, nor null out the fields that are no longer to be used.

It's Unity's proxy objects causing it, Unity uses a proxies for native object references, so when we != null compare it, the proxy is asking the engine for the native object, which is throwing as it is only allowed form main thread and Unity doco warns that ISerializationCallbackReceiver can be caused from background threads.

I pulled your branch and have found a solution, I just don't like it. We want the OnAfterDeserialize but we want to push our followup action to mainthread. So changing the OnAfterDeserialize to this fixes it.

```csharp
        void ISerializationCallbackReceiver.OnAfterDeserialize()
        {
#if UNITY_EDITOR
            //request main editor thread run the BackCompat
            UnityEditor.EditorApplication.update += BackCompat;
#endif
        }

#if UNITY_EDITOR
        void BackCompat()
        {
            //add any dragableobject already present to list for backwards compatability
            if (draggableObject != null)
            {
                if (!draggableObjects.Contains(draggableObject))
                {
                    draggableObjects.Add(draggableObject);
                }
            }

            if (targetObject != null)
            {
                if (!targetObjects.Contains(targetObject))
                {
                    targetObjects.Add(targetObject);
                }
            }
            draggableObject = null;
            targetObject = null;
            
            //this function is only used by OnAfterDeserialize to get onto main thread via EditorApplication.update and we're done now so unsub
            UnityEditor.EditorApplication.update -= BackCompat;

        }
#endif
```

Have any better ideas?

Also, was there a reason for having the targetRef be a GameObjectVariable rather than Collider2DVariable?

Assets/Fungus/Scripts/EventHandlers/DragCompleted.cs
```diff
@@ -123,28 +174,53 @@ public virtual void OnDragExited(Draggable2D draggableObject, Collider2D targetO
...
                 // Assume that the player will have to do perform another drag and drop operation
                 // to complete the drag again. This is necessary because we don't get an OnDragExited if the
                 // draggable object is set to be inactive.
+                if(draggableRef!=null)
+                {
+                    draggableRef.Value = draggableObject.gameObject;
+                }
+                if(targetRef!=null)
+                {
+                    targetRef.Value = targetObject.gameObject;
```
stevehalliwell: In testing your scene DragComplete was not firing, seems this line should be 
```
targetRef.Value = targetCollider.gameObject;
```

Worth double checking the other scripts that use targets too

SarveshBajaj: There is no Collider2DVariable defined yet so I was using GameObjectVariable instead.

SarveshBajaj: I have fixed the wrong reference. It was a special case in drag completed since the event does not store the target collider. So I had to get it from the OnDragEntered and store it in the OnDragCompleted Class.  

The upgrade of fields does seem to work when I opened the scene in editor, does it not work in runtime? (If I open a scene from the old system in editor, it doesn't throw any errors.)

And why don't you like the solution you proposed? and secondly, I'm new to Unity so I don't have an alternate solution.

SarveshBajaj:
InvalidOperationException: EnsureRunningOnMainThread can only be called from the main thread

Hi,for the above error, I have an alternate solution using awake() and [ExecuteAlways], so should I use this? Or, should I implement your solution? PR #791

SarveshBajaj:
I have implemented the solution using awake() and [ExecuteAlways], please review once if any changes need to be made, and if your solution is better I will implement that.

stevehalliwell:
I think Awake with ExecuteAlways is going to be executed in a few less circumstances than the serialisation listener, but should have very little impact. Also, there are already a few other cases of ExecuteInEditMode within Fungus, so it fits in well. I say we go with the Awake method, I'll add a comment to them after I've merged noting the alternate. If we find more of these situations stacking up, we'll probably end up having a some kind of FungusCompatabilityHelper in the editor to make them cleaner.

PR #792, opened 2020-02-11
Added GetVariables function to Flowchart class

CG-Tespy:
Thought it'd be convenient for things such as:

- Reversing or combining values for a conditional
- Altering a bunch of Transforms' attributes at once
- Linq queries for further narrowing down results


stevehalliwell:
Sounds good. 
Let's mark it virtual and just a List<T> to match the existing Variables property.

CG-Tespy:
I've applied the requested changes to the branch

Issue #778, opened 2019-11-18
If statement should allow multiple conditions


AcademyOfF:
I often wish I could have an IF command that says `if x and y` or `if x or y`.  I'd want multiple conditions, e.g., the ability to say, `if a and b and c and d...`

I'd be happy to try to implement this myself, the only problem is I don't know how to make it backward compatible so the existing if statements still function as they did.

stevehalliwell:
If you're having to alter the underlying data being stored in the command(s), you can leave the existing data there, but put HideInInspector on it. Then implement [ISerializationCallback ](https://docs.unity3d.com/ScriptReference/ISerializationCallbackReceiver.html)and in the OnAfterDeserialize method convert the existing data into the format of the new data.

Also feel free to start a PR even if it's incomplete for ongoing feedback and assistance.

AcademyOfF:
I was unable to figure out how to do this, so I'm taking a different route.  I'm creating a Boolean Algebra command.  If you'd be interested in it, let me know and I can share it.  Either way, I'd still prefer if's with multiple conditions.

AcademyOfF:
Here's what I came up with.  I don't like it very much, but it gets the job done.  Maybe i'll give you ideas, but I doubt you'll want it directly. 

https://pastebin.com/C9zh0Kxr
https://pastebin.com/p9RMxaE3

stevehalliwell:
Thanks for sharing the snippets. They'll be useful along side the conditions, variables, and looping improvements I was looking into.

preda2or:
I would like to work on the issue. Any specific design i should keep in mind during its implementation?

stevehalliwell:
Make sure you start from dev branch, there are some changes to the internals of the control flow commands and in particular the conditional command.

The imagine the UI is going to be a mess, I think we should work through that in an early PR.

Feel free to shout in here or in an incomplete PR for help and feedback.

preda2or:
For this issue, do we need brackets
eg., 

**(exp1 or exp2) and exp3**

Or simply, we just need all expressions combined with "and, or, not" ?

And in the UI, I am planning to do the simple version inside the If expression block using multiple expression blocks joined by "or, not, and" dropdowns

AcademyOfF:
Ideally, parentheses will be allowed. Otherwise, you'd have to pick which associativity you want. When you're really programming, you don't think this way. If there is doubt about which part of this statement is going to be executed first, you wrap it in parens.

[...]

PR #796, opened 2020-02-20
basic support for multiple conditions added

preda2or:
Fixes issue #778 

created basic support for multiple conditions using logic anyOneTrue or allTrue from a list of conditions. 

- Need suggestions for UI improvements. 
- One problem i have noticed is that indenting breaks the data variable field of AnyVariableAndDataPair in the custom inspector.
- For backward compatibility setting the anyVar field to null doesn't seem to work. it seems to get reinitialized on changing the number of condition. so for now I am checking if anyVar.variable != null to skip backward compatibility.

### DIFF
--- a/Assets/Fungus/Scripts/Commands/VariableCondition.cs
+++ b/Assets/Fungus/Scripts/Commands/VariableCondition.cs
@@ -3,12 +3,16 @@
 
 using UnityEngine;
 using System.Collections.Generic;
+using System.Text; 
 
 namespace Fungus
 {
 
+    /// <summary>
+    /// class for a single condition. A list of this is used for multiple conditions.
+    /// </summary>
     [System.Serializable]
-    public class conditionExpression
+    public class ConditionExpression
     {
         [SerializeField] protected CompareOperator compareOperator;
         [SerializeField] protected AnyVariableAndDataPair anyVar;
@@ -16,8 +20,8 @@ namespace Fungus
         public virtual AnyVariableAndDataPair AnyVar { get { return anyVar; } }
         public virtual CompareOperator CompareOperator { get { return compareOperator; } }
 
-        public conditionExpression(){}
-        public conditionExpression(CompareOperator op, AnyVariableAndDataPair variablePair)
+        public ConditionExpression(){}
+        public ConditionExpression(CompareOperator op, AnyVariableAndDataPair variablePair)
         {
 
             compareOperator = op;
@@ -37,12 +41,12 @@ namespace Fungus
     public abstract class VariableCondition : Condition, ISerializationCallbackReceiver
     {       
 
-        [Tooltip("The type of comparison to be performed")]
+        [Tooltip("Selecting \"Any One True\" will result in true if atleast one of the conditions is true. Selecting \"All True\" will result in true only when all the conditions are true.")]
 
         [SerializeField] protected AnyOrAllConditions anyOrAllConditions;
         
 
-        [SerializeField] protected List<conditionExpression> conditions = new List<conditionExpression>();
+        [SerializeField] protected List<ConditionExpression> conditions = new List<ConditionExpression>();
 
 
         /// <summary>
@@ -55,12 +59,12 @@ namespace Fungus
 
             if (conditions == null)
             {
-                conditions = new List<conditionExpression>();
+                conditions = new List<ConditionExpression>();
             }
 
             if (conditions.Count == 0)
             {
-                conditions.Add(new conditionExpression());
+                conditions.Add(new ConditionExpression());
             }
         } 
 
@@ -72,7 +76,7 @@ namespace Fungus
             }
 
             bool resultAny = false, resultAll = true;
-            foreach (conditionExpression condition in conditions) 
+            foreach (ConditionExpression condition in conditions) 
             {
                 bool curResult = false;
                 if (condition.AnyVar == null) 
@@ -98,7 +102,7 @@ namespace Fungus
                 return false;
             }
 
-            foreach (conditionExpression condition in conditions) 
+            foreach (ConditionExpression condition in conditions) 
             {
                 if (condition.AnyVar == null || condition.AnyVar.variable == null)
                 {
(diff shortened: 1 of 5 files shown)


stevehalliwell:
Had a quick look over it, seems like a good start, some coding standard stuff to fix up but that's not a big deal at the moment.

Some quick suggestions to chase up.
- Use OnValidate to force list to be at least 1 long, empty condition list makes no sense
- Gaps between conditions, via Space or Separator
- RE; AnyVar not nulling, try putting [HideInInspector] on it, the serialized object may still be defaulting it even though the custom editor isn't showing it.
- Definitely do indent, the data drawerer still functions but does have a visual error, I'll see I can fix that.
- Add a demo scene somewhere in FungusExamples to demonstrate and validate functionality.
- The summary will end up being the pseudo code for the condition chain, but it may not fit easily in the inspector space, I would try putting that in a label above or below the rest of the command for users to cross check against.

If you are still wanting to extend it to have A && B || C, then having that stored in your ConditionExpression class makes sense, just don't draw that element on the last item in the list. If you are going to include that feature, you'll definitely want to look at using reorderable list for the ConditionExpression list though

preda2or:
Addressing your issues in order

1. In OnValidate if the list is empty i have added an empty condition to it.
2. Added some space between conditions.
3. I tried adding [HideInInspector] to it, but it is still not nulling. Is there a problem with checking anyVar.variable == null as i am doing right now?
4. Added the code for indenting but it's commented out for now because if i indent the data drawer has a visual glitch as well as the drop down isn't working. You can un-comment that when you fix it.
5. Added a demo scene(MultipleConditionsExample).
6. I'm not sure which summary you are referring to here. I have already changed the GetSummary() of the VariableCondition class in my previous commit. It displays next to the if or else if command in the command list window itself.

I am not planning to support conditions like A && B || C directly in the if or else if commands because that requires brackets/priority. It can be done in a separate command though. But as you said people who want that kind of behavior will probably use lua if.

Assets/Fungus/Scripts/Commands/VariableCondition.cs
```diff
@@ -6,49 +6,147 @@
 
 namespace Fungus
 {
-    public abstract class VariableCondition : Condition, ISerializationCallbackReceiver
+
+    [System.Serializable]
+    public class conditionExpression
```
stevehalliwell: Should start with Capital and could use  a class summary

Assets/Fungus/Scripts/Commands/VariableCondition.cs
```diff
@@ -6,49 +6,147 @@
...
+    // anyone with a better name for this can update it
+    public enum AnyOrAllConditions
+    {
+        AnyOneTrue,
+        AllTrue
+    }
+    public abstract class VariableCondition : Condition, ISerializationCallbackReceiver
+    {       
+
+        [Tooltip("The type of comparison to be performed")]
```
stevehalliwell: Prefer explaining what the two modes mean/ what they will do.

Assets/Fungus/Scripts/Commands/VariableCondition.cs
```diff
@@ -6,49 +6,147 @@
...
-            summary += anyVar.GetDataDescription();
+            string summary = "";
+            string connector = "";
+            if (anyOrAllConditions == AnyOrAllConditions.AnyOneTrue)
+            {
+                connector = " Or ";
+            }
+            else
+            {
+                connector = " And ";
```
stevehalliwell: These don't read easily in the command summary preview, try all caps and maybe wrap in bold tags

Assets/Fungus/Scripts/Editor/VariableConditionEditor.cs
```diff
@@ -33,61 +33,90 @@ public class VariableConditionEditor : CommandEditor
...
 
-            if (selectedVariable != null)
-            {
-                compareOperatorProp.enumValueIndex = selectedIndex;
+                GUILayout.Label("Condition "+(i+1).ToString(),EditorStyles.boldLabel);
+
+
+
+                //uncomment this for indentation
+                // EditorGUI.indentLevel++;
```
stevehalliwell: Pushed a fix for the data drawer to develop

Assets/Fungus/Scripts/Editor/VariableConditionEditor.cs
```diff
@@ -33,61 +33,90 @@ public class VariableConditionEditor : CommandEditor
...
-
-            selectedIndex = EditorGUILayout.Popup(
-                new GUIContent("Compare", "The comparison operator to use when comparing values"),
-                selectedIndex,
-                operatorsList);
 
-            if (selectedVariable != null)
-            {
-                compareOperatorProp.enumValueIndex = selectedIndex;
+                GUILayout.Label("Condition "+(i+1).ToString(),EditorStyles.boldLabel);
```
stevehalliwell: Some kind of heading before the condition list would be a good, but a heading per condition seems overkill. Ends up taking a lot of space.

Assets/Fungus/Scripts/Commands/VariableCondition.cs
```diff
@@ -6,49 +6,147 @@
...
-            if (anyVar.variable == null)
+            if (!this.HasNeededProperties())
             {
                 return "Error: No variable selected";
             }
 
-            string summary = anyVar.variable.Key + " ";
-            summary += VariableUtil.GetCompareOperatorDescription(compareOperator) + " ";
-            summary += anyVar.GetDataDescription();
+            string summary = "";
```
stevehalliwell: Given how many += end up being done here and given following loop. It's probably worth moving this to a StringBuilder.

preda2or:
Made your requested changes

stevehalliwell:
Checked out locally to test out, found a problem in back compat. When updating a command from old fields to new version, it's ending up adding it multiple of the same condition. Due to not nulling out the existing old data after moving to new.

stevehalliwell:
I'm going to pull into a local branch and make the required adjustments before merging into develop.

PR #800, opened 2020-02-25
Heirarchy Icon placement bug in newer Unity versions

TonyJeffree:
The removed line was not functioning as intended, it was covering the disclosure triangle of a parent GO with the mushroom icon if that parent GO also had a flowchart component.

I'm sure there was a reason for this logic change, so I've left the if cases intact, but this was necessary in my own project so I thought I'd share it.

Tested locally and works fine. This is my first ever community git contribution, so apologies if I've committed a faux pa!

### DIFF
--- a/Assets/Fungus/Scripts/Editor/HierarchyIcons.cs
+++ b/Assets/Fungus/Scripts/Editor/HierarchyIcons.cs
@@ -70,7 +70,7 @@ namespace Fungus
             // place the icon to the left of the element
             Rect r = new Rect(selectionRect);
 #if UNITY_2019_1_OR_NEWER
-            r.x -= r.height;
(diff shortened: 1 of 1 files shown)


stevehalliwell:
Thanks for contributing.

As I recall there was a unity version where the rect returned for hierarchy items was giving a subset of the row space. I'll double check against a few Unity versions before merging.

stevehalliwell:
Ran through a number of unity versions and I was remembering correctly, zero there will overlap the visibility icon. It does need some x offset just more than r.height so it doesn't cover the foldout icon on newer versions.

stevehalliwell:
I'm going to pull into a local branch and make the required adjustments before merging into develop.

PR #849, opened 2020-07-04
Fixed massive slowdown in editor when there is MANY objects in the scene

msklywenn:
### Description
I just got into a project that had the editor run VERY slow. I noticed there was thousands of calls to DrawGizmo even if there wasn't much gizmos in the scene. I narrowed this down to Fungus.EditorUtils.ViewEditor.RenderCustomGizmo which was really surprising because Fungus was not used in the scene I was working on.

Has it turns out, the DrawGizmo attribute was being called for ALL objects in the scene and there is more than ten thousand objects in this scene so it was crippling the performance of the editor. This was easy to fix: just filter out all the objects that don't have the View component by using the second constructor of the DrawGizmo attribute.

Editor performance being crippled if there is MANY (10000+) objects in a scene even if Fungus is not being used in that scene.

### What is the new behavior?
Normal performance. (0 calls to Fungus.EditorUtils.ViewEditor.RenderCustomGizmo if there is no Fungus View gizmo to render)

### DIFF
--- a/Assets/Fungus/Scripts/Editor/ViewEditor.cs
+++ b/Assets/Fungus/Scripts/Editor/ViewEditor.cs
@@ -18,17 +18,14 @@ namespace Fungus.EditorUtils
 
         // Draw Views when they're not selected
 #if UNITY_5_0
-        [DrawGizmo(GizmoType.NotSelected | GizmoType.SelectedOrChild)]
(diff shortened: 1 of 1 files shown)


stevehalliwell:
Thanks very much for the fix. :)

Issue #838, opened 2020-06-10
Narrative log does not show empty say commands.


Vaqq:
**Is your feature request related to a problem? Please describe.**
I made a emotion system, for example character says "Hello", then block contains three Say commands(Eyes, emotions, mouth). Only eyes Say command contains text "Hello", so narrative log only displays said text, but for empty Say commands it displays empty space.

**Describe the solution you'd like**
I would like that if Say command is empty(contains no text), then narrative log will not display an empty space in the log. It will make narrative log long and fill it with empty space, not a convenient way for the player to scroll and scroll looking for the desired dialog.

**Additional context**
![Say_Commands](https://user-images.githubusercontent.com/66467367/84252463-477a0d00-ab17-11ea-9f18-4319336c3f5c.jpg)

![Narrative_Log_Say_Commands](https://user-images.githubusercontent.com/66467367/84252546-637dae80-ab17-11ea-85e7-20114e790894.jpg)

Thanx!

CG-Tespy:
Could you create a mockup for what you want the end result to look like?

PR #866, opened 2020-09-08
Added requested feature: Narrative Log not registering empty Say dialogs

CG-Tespy:
### Description
A new feature: the option to have the NarrativeLog not log empty Story texts

The narrative log logs all instances of Say Dialogue story text, regardless of their contents. Relevant issue: #838

### What is the new behavior?
You can have (as the issue asks) it so Narrative Logs ignore all empty Say Dialogue text. All you have to do is add the new RemoveEmptyNarrativeLogEntries to the NarrativeLog prefab. 

### Important Notes
- To help make this work, I added a new command: Clear Narrative Log. Calling the NarrativeLogEntryUI's Clear method only clears the UI, not the underlying log, which can cause issues with this new feature... This command would prevent that, and should thus be the default way users clear Narrative Logs. 
- I've also refactored other parts of the existing narrative system for this.
- My changes require modifications or additions to documentation
- My changes modify the workflow/editing of the main flowchart in NarrativeEntryLogTest
- My changes modify the runtime execution/behaviour of the narrative system
- My changes modify the existing NarrativeLogEntryTest scene


stevehalliwell:
Looked at this a few times. 

I need to make some tests to confirm my thinking before making premature suggestions.

CG-Tespy:
It's better that I have this feature implemented through a separate branch

PR #870, opened 2020-09-15
Portrait Move To Front Fixed

### LINKED ISSUE #869: Move To Front does not work
Portrait - MoveToFront does not work.

**To Reproduce**
Add two characters (c1, c2) and `Show` c1, then c2 at same stage position (e.g. Left). Use `Portrait - MoveToFront` block to move c1 to the front.

**Result**
c1 remains in the background.
c2 is in the front.

**Expected behavior**
c1 is in the front

**Screenshots**
characters in stage canvas
While the game is playing, the gameObject `Stage` > `Canvas` has character holders as children. These character holders have portraits as children.

The current code rearranges the children of the character holder objects.
Instead the code should rearrange the character holders.

**Versions & Platform (please complete the following information):**
 - Unity version 2020.1.2f1
 - Fungus Version 3.13.4
 - Editor OS: Win10
 - Build Platform: Occurs in editor and std win build.

iwanPlays:
Fixes #869 

- If adding prefabs or scenes, please make them in the oldest supported version of Unity possible (to ensure they are able to run in the Compatability release)
I used 2020 to create a scene but it does open in 2019.4.8f1.
2018.4.26 doesn't open the project without error (something about packages) so maybe that's not supported anyways? Can't find documentation on which unity versions are supported by Fungus.

### Description
Bugfix for #869 

Move To Front does nothing

### What is the new behavior?
Move To Front moves character's portraits parent to the front

### Important Notes

- My change adds 1 test scene

### DIFF
--- a/Assets/Fungus/Scripts/Components/PortraitController.cs
+++ b/Assets/Fungus/Scripts/Components/PortraitController.cs
@@ -390,7 +390,7 @@ namespace Fungus
         /// </summary>
         public virtual void MoveToFront(PortraitOptions options)
         {
-            options.character.State.portraitImage.transform.SetSiblingIndex(options.character.State.portraitImage.transform.parent.childCount);
+            options.character.State.portraitImage.transform.parent.SetSiblingIndex(options.character.State.portraitImage.transform.parent.parent.childCount);
             options.character.State.display = DisplayType.MoveToFront;
             FinishCommand(options);
         }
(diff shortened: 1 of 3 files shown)


breadnone:
Confirmed that it's broken. Did not try the solution on this pull tho

stevehalliwell:
Thanks for the report and the PR. 

After merging in, I think I'll modify the change to address the holder directly rather than the image parent transform, incase further heirarchy changes are made to portraits down the line.

As for versions supported, there are branches previous Unity versions, as opening a project in an older version of Unity will often result in errors or unreable data. The most common one now is packages in the manifest being incompatible.

PR #875, opened 2020-09-19
Add Simple Tint preferences

### LINKED ISSUE #868: A way to Change Color of Commands
**Is your feature request related to a problem? Please describe.**
Some colors schemes can hit individuals differently. On some days, the If/Loop/etc commands color (yellow) shows up as an irritant for light sensitivity. While Wolfrug is correct I could change it myself, this would create a problem when Fungus updates, so that is not a practical idea.

**Describe the solution you'd like**
One idea would be create a few color schemes to select from the Tools>Fungus window, so users can quickly change to that even after updating.

stevehalliwell:
CommandListAdapter items and Blocks in flowchartwindow have configurable tints in fungus editor preferences.

image

### Description
Adding simple tinting for elements of Fungus UI as per close #868 

### What is the new behavior?
In the Fungus Editor Preferences the user can now configure a tint colour for the CommandList in the Block Inspector and a tint for the Block icons in the Flowchart Window.

### Other information
These type of tints are easy to integrate and should remain easy to maintain. I'll wait for feedback before mering this PR in case there are more elements that are requested.

### DIFF
--- a/Assets/Fungus/Scripts/Editor/CommandListAdaptor.cs
+++ b/Assets/Fungus/Scripts/Editor/CommandListAdaptor.cs
@@ -162,6 +162,8 @@ namespace Fungus.EditorUtils
                     break;
                 }
             }
+            var prevCol = GUI.color;
+            GUI.color = FungusEditorPreferences.commandListTint;
 
             string commandName = commandInfoAttr.CommandName;
             
@@ -382,6 +384,7 @@ namespace Fungus.EditorUtils
             GUI.Label(summaryRect, summary, summaryStyle);
             
             GUI.backgroundColor = Color.white;
+            GUI.color = prevCol;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


danac78:
So if I understanding this right, this will change the tint for commands (say change the If command from yellow to a color of my choice)? If so, I cannot think of anything else. Not sure where the Fungus box is though where you are setting the tints in the screenshot.:)

stevehalliwell:
@danac78 It's less targeted than that, it's allowing a global tint to all the colours in the Block inspector and in the Flowchart window. For the use case described in the issue, I thought this might be a good stop gap at least. Being able to turn down the intensity of the colours, the tint is just a multiply to the colour given by GetButtonColour.

danac78:
@stevehalliwell  Ok, I see where you are going with it. In that case, then yes, it is good. :)

Issue #879, opened 2020-09-23
QoL: Ability to cycle through commands in the Inspector


TheEmbracedOne:
**Is your feature request related to a problem? Please describe.**
It would be super useful if there was a quicker way to cycle through commands on a block. For example if I want to double check all Say commands have the correct character name, it'd be handy to have a faster way to cycle through each command in the current block.

**Describe the solution you'd like**
When pressing up/down keys, I'd like the next/previous command selected, displaying its content below like on a regular click; like this, but on up/down key presses on the keyboard instead of clicking:
![zBgnjKE9OQ](https://user-images.githubusercontent.com/16197906/94043218-1b2f8700-fdbc-11ea-8c68-f57aea9a9e64.gif)

This works in the hierarchy window of Unity by default:
![AkDATq2lje](https://user-images.githubusercontent.com/16197906/94043311-46b27180-fdbc-11ea-9390-6bae9032faec.gif)

Vaqq:
I'm using "page up" and "page down" on keyboard, to cycle through commands quickly.

TheEmbracedOne:
Wow @Vaqq that actually works! This piece of code actually resides in Fungus\Scripts\BlockEditor.cs, marked with comments `// Next Command` and `//Previous Command`. Modifying that bit of the code from this:

```
            // Previous Command
            if ((Event.current.type == EventType.KeyDown) && (Event.current.keyCode == KeyCode.PageUp))
            {
                SelectPrevious();
                GUI.FocusControl("dummycontrol");
                Event.current.Use();
            }
            // Next Command
            if ((Event.current.type == EventType.KeyDown) && (Event.current.keyCode == KeyCode.PageDown))
            {
                SelectNext();
                GUI.FocusControl("dummycontrol");
                Event.current.Use();
            }
```

to this:
```
            // Previous Command
            if ((Event.current.type == EventType.KeyDown) && (Event.current.keyCode == KeyCode.PageUp) ||
                (Event.current.type == EventType.KeyDown) && (Event.current.keyCode == KeyCode.UpArrow)) {
                SelectPrevious();
                GUI.FocusControl("dummycontrol");
                Event.current.Use();
            }
            // Next Command
            if ((Event.current.type == EventType.KeyDown) && (Event.current.keyCode == KeyCode.PageDown) ||
                (Event.current.type == EventType.KeyDown) && (Event.current.keyCode == KeyCode.DownArrow)) {
                SelectNext();
[...]

PR #880, opened 2020-09-24
Adding ability to cycle block commands with up/down keys in addition to existing pagedown/pageup key navigation

TheEmbracedOne:
#711 # Description
Added ability to cycle Editor block commands with up/down arrow (already possible with pagedown/pageup)
Related to #879 

Navigating block commands in the editor is only possible with PageUp and PageDown keys.

### What is the new behavior?
Navigating block commands in the editor is only possible with PageUp and PageDown keys AND Up and Down arrow keys.

### Important Notes


TheEmbracedOne:
This was a test, please ignore it

PR #881, opened 2020-09-24
Update BlockEditor.cs to support up/down key navigation

TheEmbracedOne:
### Description
Added ability to cycle Editor block commands with up/down arrow (already possible with pagedown/pageup)
Related to #879 

Navigating block commands in the editor is only possible with PageUp and PageDown keys.

### What is the new behavior?
Navigating block commands in the editor is only possible with PageUp and PageDown keys AND Up and Down arrow keys.

### Important Notes

### DIFF
--- a/Assets/Fungus/Scripts/Editor/BlockEditor.cs
+++ b/Assets/Fungus/Scripts/Editor/BlockEditor.cs
@@ -315,14 +315,16 @@ namespace Fungus.EditorUtils
 
 
             // Previous Command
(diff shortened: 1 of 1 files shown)


stevehalliwell:
Thanks for the contribution. 

I haven't got to test it yet but will do soon. Providing it's all good, I'll merge into develop. :)

I'll add the ability to turn these off via a FungusPreference once it's in develop. Just in case some users have other actions happening on key up and down in their projects.

Issue #877, opened 2020-09-22
Flexibility to control over Non-Active character


breadnone:
**Expected behavior**
Command to control over the Non-Active character. Feature to control the Non-Active/Non-Speaking character only available through Stage/ControlStage through "Dim" property and that's it.

More often than not, Non-Active/Non-Speaking character also needs to be animated which corresponds to when the Active character is currently speaking (see Additional context below). Functionality is quite similar to what ControlStage has, but with it's own twist.

_(mockup sample)_
![fungus-mockup-01](https://user-images.githubusercontent.com/64100867/93865473-91ff4e00-fcf0-11ea-91bb-c2348f01541e.png)

**GIF Screenshots**
 Note: The sample image below was hardcoded, with the intention to show how it looks like and a bit glitchy. Notice at the very first/beginning Sherlock is talking while John is bouncing left and right.
![ctgwrLgHSs](https://user-images.githubusercontent.com/64100867/93862434-4185f180-fcec-11ea-9d42-1b87ce15e488.gif)

**Additional context**
Imagine if there's one scene where the currently Speaking character" is mad/angry/sad and the "Non-Speaking" character would react to that in form of animation such as; wobbly animation, jump, bouncing left and right animation, etc...

Not only this would give freedom to the creator it would also open new possibility/creativity on how to make an interesting interactive game.

Thanks for this beautiful piece of software

stevehalliwell:
I'm very keen for this and if you'd like to keep working on it that'd be great and much appreciated by all. 

Some things I'd consider;
 - There may well be more than 1 command that interacts with the underlying logic.
  - One such command would be AnimatePortrait, that can configure and trigger the motion regardless of Active/Speaking-ness of the Character.
 - Does the animation target the portrait image or the character portrait root? I can see there being a need for both. 
   - Does there need to another transform added so there's Root->AnimRoot->PortraitImages
 - How can users add their own tweens (or animations) to the system.
   - Preferably without having to make modifications to any scripts inside Fungus asmdef when doing so.
 - Can (or should) this have support for motion driven by Animations, not just tweens.
 - Can/How does the user customise these per character.
 - How can these be trigged in Conversations.
 - How can these be trigged within text of a Say.
 - The bulk of the logic for making this happen probably belongs inside PortraitController.

I'm happy to give feedback on development in a WIP PR. If you'd like to keen working on this, let us know and I'll assign this issue to you. So no one else tries to do it simultaneously.

[...]

PR #883, opened 2020-09-26
Command to control over Active/Non-Active character on stage (WIP)

breadnone:
### Description
Custom command for flexibility/freedom to control speaking/non-speaking character.

Currently the way to control character individually is very limited especially the non-active ones. There are ways to address this partially but it requires multiple blocks and Stages, and this PR/custom command is not.

### What is the new behavior?
Add animation to active/non-active and all character on stage based on predefined tween presets.
- Add predefined tween animation to speaking character.
- Add predefined tween animation to non-speaking character.
- Add predefined tween animation to all character on stage.

### Important Notes
- My change adds new custom command

**Workflow of this custom command:**

Command name: Animation/Character Expression

1. It will loop until next character's portrait change so the extra block to stop the tween may not be required
2. To stop the tween just drop the command again and select "None" for single character and uncheck "All char" if in the previous block the All char is checked
3. (Optional) It also compatible with "LeanTween/Stop Tweens" to stop the tweening objects

Please do keep in mind, the idea of this custom command originally to make my game making process simpler so it may not suitable for you. Working with 600+k of words with thousand of commands in one scene is not really convenient and corrections can be HELL! This pr partially address that issue (at least for me).

### Other information
Related to https://github.com/snozbot/fungus/issues/877

Edit: The tween would sometimes stop if user clicks rapidly and I've no idea how to fix this

I may forgot a couple of things that was added, let me know if there are some errors.

### DIFF
--- a/Assets/Fungus/Scripts/Commands/CharacterActiveState.cs
+++ b/Assets/Fungus/Scripts/Commands/CharacterActiveState.cs
@@ -274,7 +274,6 @@ namespace Fungus
                         if (LeanTween.isTweening(c.State.portraitImage.gameObject))
                         {
                             LeanTween.cancel(c.State.portraitImage.gameObject, true);
-                            Continue();
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


breadnone:
This is more broken that I originally thought. Closing this garbage :)

PR #887, opened 2020-09-29
Add Punch/Wobble effect to speaking character

breadnone:
Add Punch effect to the speaking character, this can be used together with Dim as well.

The idea came from our tester, as he can't tell who's actually speaking if "Dim" wasn't being used (even though the name of the character was displayed)

### Description
New added feature as shown in the gif below:
uImpjeIwEC

Add new checkbox in Stage setting
flashportraitFungame

This feature is not explicitly available in Fungus

### What is the new behavior?
Adding new Punch tween setting to be part of Fungus's Stage system including ControlStage

### Important Notes
- Add new public bool to PortraitUtils
If this breaks Fungus code standard or if it's not allowed, pls let me know, 
- The name "Flash" displayed in Stage setting because the word "Punch" would confuse users. Obviously, this can be changed.

### Other information
It should not break compatibility.

### DIFF
--- a/Assets/Fungus/Scripts/Commands/ControlStage.cs
+++ b/Assets/Fungus/Scripts/Commands/ControlStage.cs
@@ -23,7 +23,11 @@ namespace Fungus
         /// <summary> Undim all portraits on the stage. </summary>
         UndimAllPortraits,
         /// <summary> Dim all non-speaking portraits on the stage. </summary>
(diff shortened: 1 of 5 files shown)


breadnone:
I really think there should be combobox to choose what effects would be used rather than check boxes. Not only it would make things more tidy but also make new addition for character effects easier.

But that should be in separate pr...

PR #890, opened 2020-10-02
Auto language font change

CG-Tespy:
### Description
A new feature: making TMProUGUI text fields automatically change their fonts based on what the current active language is. It includes small refactors to the Localization class so it can work, along with other new code.

There is nothing to make the fonts change based on the currently active language, taking away style opportunities for localization.

### What is the new behavior?

Whenever the Localization instance's ActiveLanguage property changes, an event is fired, alerting listeners that the active language has changed. The new LocalizationFontSyncer component responds to that change by seeing if the language has any font assigned to it (as per its Font Links field in the Inspector), and changes things accordingly. If the new language has no font assigned to it, the component opts for the default the user sets in the Inspector.

### Important Notes
- Added a new class called LocalizationSignals that works like the other signal classes, but is for listening to what happens in the Localization system
- Gave the Localization class's ActiveLanguage property a setter so it can fire off the event
- Made the Localization class's SetActiveLanguage function call the ActiveLanguage setter. This may count as a bug-fix
- There were originally some changes to the Set Language class, but as I got further into this feature's development, I realized it was fine to leave it as-is. Please feel free to ignore the changes this commit applies to that class, as well as some other files I didn't really mean to change.
- I added some fonts to this commit to help test out the feature. I'm not suggesting you add them to any official release; they're just for testing.
- My change adds a FungusExample under Localization/LanguageFontSyncing. The LocalizationFontSyncer is attached to the SayDialog's NameText and StoryText GameObjects; you may want to play around with them.
- I would've added the FontSyncer to the SayDialog prefab, but I wasn't sure what fonts you would like. Feel free to let me know which, so I can modify the prefab accordingly
- My change modifies the runtime execution/behaviour of the Localization class


CG-Tespy:
I redid how FontLinks work; the users are to create FontLinkSet Scriptable Objects and assign those to the FontSyncer instances. This way, users won't have to mess around in the inspector as much when they want to change font-language associations for different text fields.

stevehalliwell:
Cool addition.
Quick bits:
 - I'm not going to try and fit this into 3.14, still a lot to do there and it needs to stabalise. 
 - Make sure any included fonts have their licenses next to them.
 - This needs a Non-TMPro text solution too.

CG-Tespy:
> Cool addition.
> Quick bits:
> 
> * I'm not going to try and fit this into 3.14, still a lot to do there and it needs to stabalise.
> * Make sure any included fonts have their licenses next to them.
> * This needs a Non-TMPro text solution too.

Understandable, and I'll have a regular Unity Text solution added to this in the future

CG-Tespy:
I've added said support now ^_^ Overview of the changes:

- Switched out Hey Gorgeous since the original creators no longer seem to be around; I can't find the license anywhere
- Added a couple other fonts, making sure to include their licenses
- Added a FontAdapter class that can work with both text systems' font types
- Refactored the LangFontLink classes, and then the FontSyncer to work with those changes
- Added to the TextAdapter class so that you can change text fields' fonts through a FontAdapter, making it easier for the FontSyncer to do its thing
- The test scene lets you choose whether you want to see text in a TMPro or UGUI textbox, so you can make sure that things work for both text systems

There's an awful lot to take in, here xD

Issue #806, opened 2020-03-12
"Wait before next Say dialogue" option


TheEmbracedOne:
Hi, I'd like to request the addition of a public variable to Say or Writer so that users could configure how long a say command should wait after finishing before showing the next say command.

I have "wait for click" off, and right now if a say dialogue consists of a 1-2 word sentence, it is on the screen for less than a second. Just like I'm able to specify punctuation pause, it'd be great to be able to specify a minimum duration for how long each say dialogue should be displayed for.

There's probably a abetter way to do this, right now I was able to get this to work the following way:

Say.cs script, from line 133
```
sayDialog.Say(subbedText, !extendPrevious, waitForClick, fadeWhenDone, stopVoiceover, waitForVO, voiceOverClip, delegate {
                StartCoroutine(ExampleCoroutine());
            });
        }

        IEnumerator ExampleCoroutine()
        {
            yield return new WaitForSeconds(waitForDuration);
            Continue();
        }
```

Hope you'll consider adding this!

TheEmbracedOne:
PS. the problem with my implementation is that while I got the say editor to display this field and I can give it a value, I'd like to have this as a global value, ie on the Writer script or someplace similar, so I wouldnt have to set it for every single say dialogue.

TheEmbracedOne:
apparently what I wrote above does not work as intended and to clarify, the intent here would be to make sure a piece of text stays on the screen for a desired X secs before the next Say command triggers.

stevehalliwell:
Writers currently support a wait token which will hold for a given number of seconds. Such as {w=1.5}. It seems to be intended for author to configure dramatic delays as required. Given you want a global one, I would recommend adding some code to your writer or say that appends a string with your desired wait time to everything it is asked to process. This is similar to how wait for input is handled presently.

TheEmbracedOne:
Hi @stevehalliwell and thanks for the reply. I think I know which wait token you mean (punctuation pause?), but what I'd like to do is ensure that each Say command stays on screen after finishing writing for X duration. I tried adding a Coroutine as per my above example, but unfortunately I cannot figure out where in Say.cs or Writer.cs I should add this, or any other code, because I'm not a programmer. Could you advise me on this please? I'd greatly appreciate it.

CG-Tespy:
Would something that effectively adds a {w=insertUserDesignatedNumberHere} at the end of every Say command's story text be good enough, or...?

TheEmbracedOne:
[...]

PR #895, opened 2020-10-04
New feature: Say End Delay

CG-Tespy:
### Description
New feature: a delay applied at the end of each Say command (0 seconds by default), which the user can alter with a new command.

If the user wanted to have a lot of their Say commands have a delay at the end, they'd have to manually add wait tags to them. Relevant issue: #806

### What is the new behavior?
This new feature adds the wait commands for the user behind the scenes, with a duration they can set at their leisure. using the new Narrative command: Set Say End Delay.

### Important Notes
- My change modifies the runtime execution/behaviour of the Say command.
- My change adds a test scene to the FungusExamples, in the folder named "SayEndDelay", so you can see the feature in action.
- The invisible wait-tag-adding is done through changes in the Say.cs file. 
    - I've also done a significant readability refactor of its code; I figured I might as well.
        - Of course, I ran some of the other Say-related FungusExamples to make extra sure I didn't break anything.

### Other information:

This was a very easy feature to implement, and it'd be just as easy to implement a delay at the beginning of a Say command. Perhaps you can poll the community, asking if they'd want that?


stevehalliwell:
I'll have to merge this in manually/cherry pick as it appears to be partially work from another PR and some files that shouldn't have made it through the gitignore.

Issue #882, opened 2020-09-26
Debug Log for Invoke Method/Invoke Event errors


danac78:
**Is your feature request related to a problem? Please describe.**
One thing that would be helpful is a debug log to indicate where the Invoke Method is when it is causing an error (null, empty, incompatible.) The last one usually occurs when something changes and I had to redo the save system so I know where those are. However, I use a lot of invoke methods to take care of addressable. Now, checking 2 or 3 in the same block isn't a problem, but when there is 10-14 blocks in a flowchart it becomes a long hunt for it.

**Describe the solution you'd like**
Similar to Lua when you use Execute Lua and write code: If it throws an error, it tells you the name of the block it is in. However, something similar is not available with Invoke Event or Invoke Method other than telling me what line in InvokeMethod.cs that throwing the error. By giving a clue which block (especially since we should use unique names for them), it isolates the problem faster.

stevehalliwell:
Can certainly add try catches around the offending lines in this command. I think this also prompts a more general solution too. Perhaps Block is too trusting, could put a try catch in there and at least log the unique identifying location of the command that threw and then rethrow.

danac78:
That sounds good.

TheEmbracedOne:
I had a problem with tracking down 2 Invoke Method-related errors the other day in a flowchart with over 2000 commands across hundreds of blocks. What I ended up doing was the following:

(right above `public override Color GetButtonColor()`):
```
            } catch (System.Exception ex) {
                Debug.LogError("Error: " + ex.Message + "block: " + ParentBlock.BlockName);
            }
        }
```
The above Debug.Log helped me find what block the error was coming from. Your since-submitted solution is definitely better, but I still wanted to share my findings regarding this. Perhaps it might help someone in the future approaching another issue.

stevehalliwell:
@TheEmbracedOne no, you're right #889 won't do anything about errors from awake. But that should show as an error in the command list, Invoke checks targetobject in its summary.

TheEmbracedOne:
@stevehalliwell I see, thanks for the clarification! Would it be possible to add something to the errors from awake, to help track down what block they're coming from please? My "solution" above worked for a specific scenario only, and I'm sure someone more experienced will know of a better way to do this.

danac78:
Seem like the solution resolving this has been merged..;)

PR #901, opened 2020-10-09
Add info to logerror for InvokeMethod Awake

stevehalliwell:
Remove bespoke exception handling in cases where block command catch can do it better.

### Description
See #882 

The Awake tries to catch and logbut its info is not overly helpful.

### What is the new behavior?

- the awake and onenter of the InvokeMethod will now both result in a rethrow and a logerror with the location information
- other commands with custom try catch inside OnEnter have been removed in favour of calling block handling it.


danac78:
@stevehalliwell Ok, I added the code because I was running into an invoke method error and had a devil finding it. Applied all from the previous and this commit, and found where it is with more ease. Thanks 👍

Issue #904, opened 2020-10-12
Explosing CommandIndex and OnComplete for StartExecution on Block.cs


MrMMatricks:
Hello!

It would be great to have in the Block class the parameters from
`Execute(int commandIndex = 0, Action onComplete = null)`

to be settable in StartExecution as well. 
```
public virtual void StartExecution(int commandIndex = 0, Action onComplete = null)
{
	StartCoroutine(Execute(commandIndex, onComplete));
}
```

TheEmbracedOne:
@MrMMatricks whats the use case for this? Just so I can understand your request better

MrMMatricks:
The same as for 
`Execute(int commandIndex = 0, Action onComplete = null)`

I want to start the execution of a Block from the outside and not be responsible for the handling of the Coroutine. (And be able to jump to a specific commandIndex).

We need this for somthing like a reset system where be jump back to the last sucessfull command in the block when the player fails from a Scipt that is not part of fungus.

TheEmbracedOne:
Ah, thank you, I understand now. I think it might be related to #830 
Is this something you started experimenting with, maybe trying to customise the source code? If you tried, it might help posting your findings.

MrMMatricks:
Yes, it's related. I want to control Fungus "from the outside".

And yes I would change the source of fungus and I already posted the change up there. I would change in Block.cs line 208  from:

```
public virtual void StartExecution()
{
      StartCoroutine(Execute());
}
```

to

```
public virtual void StartExecution(int commandIndex = 0, Action onComplete = null)
{
	StartCoroutine(Execute(commandIndex, onComplete));
}
```

By providing the optional parameters the interface stays the same and is backwards compartibel. So no breaking changes and nearly no sideeffects.

Cheers!

PS:
I could do a pull-request if that helps? I just never did one.. So bear with me on that one :D

TheEmbracedOne:
Nice one. Yeah pull request definitely helps, I think it speeds up Steve reviewing your request/idea.
I did my first one just the other week as well :)

PR #906, opened 2020-10-15
Add optional parameters to StartExecute in Block

MrMMatricks:
### Description
This is used to control a Flowchart from the outside. Including jumping to a specific CommandIndex or getting a callback.
This was possible before by directly starting a Couroutine wrapping Execute with the parameters but then you have to maintain the Coroutine yourself.

There are no parameters on the StartExecute Method in Block.cs. So we can't start a block from the Outside and jump to a specific CommandIndex or get a callback when the block is done.
Issue Number: #904 

### What is the new behavior?
Added optional Parameters to StartExecute so that we can jump directly to a specific CommandIndex and get a Callback when the block execution is done. 

### Important Notes

### DIFF
--- a/Assets/Fungus/Scripts/Components/Block.cs
+++ b/Assets/Fungus/Scripts/Components/Block.cs
@@ -205,9 +205,9 @@ namespace Fungus
         /// <summary>
         /// Start a coroutine which executes all commands in the Block. Only one running instance of each Block is permitted.
         /// </summary>
-        public virtual void StartExecution()
+        public virtual void StartExecution(int commandIndex = 0, Action onComplete = null)
         {
-            StartCoroutine(Execute());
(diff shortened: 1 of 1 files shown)


breadnone:
You can simply pass the index number through Continue method without making new ones  
`Continue(CommandIndex + yourIntHere);`

MrMMatricks:
That's nice to know!
But does not help in this case. I am not in a Command or EventHandler. 
I am trying to start a Block from "the Outside"

stevehalliwell:
By adding default params to StartExecution we break the interface users of Fungus may rely on (inher or bind from UnityEvent etc.)

Execute itself is public so it can be called directly those params are desired by your code. If calling as block.StartCoroutine(block.Execute(0, null)); is undesirable then I'd prefer adding an overload or new function perhaps StartExecutionExt that takes the params.

Issue #929, opened 2020-12-15
Feature Request: WriterAudio Fade Out Disabler


SoulKarl:
Feature request for a bool to disable the volume lerping in WriterAudio that occurs during text writing whenever there is punctuation.  Currently there is a dopler effect on everything and when adding in my own audio it makes it impossible to predict/control.  Having a way to disable this would be great.

Arylos07:
In PR #934, I added a toggle to the WriterAudio script that disables the audio lerping. This is left enabled by default so you'll have to have your own say dialog prefab that toggles this or change the default Fungus say dialog prefab. 

I did not add a Fungus command for this, but extending the Say command to toggle this as well could be useful, but I think that's better left to creating a say dialog for any characters that need lerping disabled for their voices.

PR #934, opened 2020-12-28
Added Requested Feature: Writer Audio lerp toggle

Arylos07:
### Description
Added a toggle to Writer Audio so the Writer does not lerp audio while writing. This results in the volume being consistent while writing. 

Default and prefab Say Dialogs have this enabled so they will lerp audio by default. The user must change the prefab settings or create a custom Say Dialog to make use of this change.

Issue Number: #929 

When using a sound such as typewriters or other effects, the lerping audio can cause the sound to change volume rapidly and not sound consistent. This can be replicated by doing the following:

- Open the WritingSpeedTest scene
- Adjust the writers on the SayDialogs
     - Change beep sounds to contain only the MidVoice sound effect.
- Play the scene.

The sound modulates heavily as it stops and stops. With the other voices, it's not heard, but with one consistent sound effect, the effect is noticeable and causes peaks in the audio levels.

### What is the new behavior?

- When Lerp Writer Volume is enabled, the default lerp volume behaviour remains
- When Lerp Writer Volume is disabled, the writer remains at the target volume set by the Writer without lerping. 

### Important Notes

     - Possibly, the variable may need to be added as well as a description of what it does. There is a tooltip in the editor that is thorough.

### Other information

Setting in editor


Arylos07:
Closing this for now. Made a fresh develop branch and opening a new PR using that fresh and clean branch.

PR #942, opened 2021-01-11
[Bugfix] Addressing Draggable2D UI Lag

CG-Tespy:
### Description
Bug fix

When you try to drag a UI element with a Draggable2D component, it lags behind, moving so slowly.

### What is the new behavior?
Dragging UI elements works as well as dragging regular 2D sprites does with the Draggable2D component

### Important Notes
- My change modifies the runtime execution/behaviour of the Draggable2D component

### Other information
- The bug came about because of how ScreenToWorldPoint doesn't play nice with how Canvases manage space. So I thought it'd be easier to have the position-updating work off local position instead of world position
- To go with the above, the newPosition variable has been renamed to "newLocalPosition"
- For better naming, I renamed the delta variable to "mouseMovement"
- For optimization, I had the main camera cached in Awake
- To make the component work with both UI images and Sprites, I had to incorporate two algorithms in this script. So we don't have to use if statements, I set up a dictionary containing them. See UpdateLocalPositionForUI and UpdateLocalPositionForSprites


CG-Tespy:
The latest bugfix made things work off of world position again. Updated the PR OP

CG-Tespy:
Just noticed another bug: when you have a Sprite using the Event System, the drag effect gets greatly exaggerated. I'll be looking into a solution

PR #943, opened 2021-01-12
Added feature: Say dialog text limits

Arylos07:
### Description
Who has had that moment of typing out a long bit of dialogue and forgot to see if it's too long? With me, **way** more than I care to admit. 

So, this PR introduces Say Dialog character limits. 

Screenshot_52

Say Dialogs now have new properties indicating the maximum text that can be displayed. These values are set by the developer based on how they design their dialogs; smaller text means larger text limits. 

- Character Name Limit: This is a limit of how much of the character name can be displayed. Any character names longer than this get an error.
- Story Text Limit: This is a limit of how much story text can be displayed in this dialog. 
- Portrait Story Text Limit: This is a Story Text Limit, but a limit if a portrait is being displayed. 
     - This value is not automatic. You have to gague how much can fit based on your average/maximum portrait size

You can set the story text limits to 0 to turn off the counter. The counter will then display "[No Story Max Length]" and will not throw errors no matter how long you make it. This really shouldn't be done, but there are some cases where having a really long dialog box would be helpful. 

Text tags do not count towards this limit, so **{b}**, **{vpunch}**, etc are not counted.

As you type character/story text, there is no feedback of how long your text will be. Based on your Say Dialog set up, this can be problematic as you either split up text too much thinking there's too much in one say command, or you put too much in and you get some text cut off. 

**Granted, you _could_ use an infinite text box or using Best Fit, but depending on your application, thay may not be the best option. **

### What is the new behavior?
- You can now adjust the text limit values for Fungus to use. Fungus will use whatever the active flowchart will be for a specific command and will display the name of the Say Dialog that command will use next to the counter.

- Anything longer than the proposed limits get an error. Say commands now also have a character counter below the story text that display this as well.

Screenshot_53
Text close to the max will make the counter yellow
Screenshot_54
Text over the limit will make the counter red and show an error in the summary.
Screenshot_58
Screenshot_55

- Text that is too long will still display in the story text (there may be reasons, after all). But when the command is run, the text is replaced by a red <!>

Screenshot_56
Screenshot_57

- Variables display a rough estimate, symbolised by "~x/y", meaning the counter is only estimating. This is because variables can change length at runtime so the value will never be 100%. It uses the default value of the variable in the flowchart to estimate this.

Screenshot_61

- Say command summaries now truncate the story text by 96 characters. Could be helpful depending on the setup.

### Backend changes
- StringFormatter has 2 helper functions added
     - Truncate (int maxLength)
          - This function truncates a string and adds "..." to the end. It does nothing to the string if it is not larget than maxLength. You can supply your own appending suffix so it doesn't _have_ to be "...".
     - SterilizeString()
          - This extension to strings clears text tags and returns only text. This function does not substitute variables. This function does not trigger an editor warning if there is an invalid tag.

- TextTagParser had a function edited
     - Tokenize now has a variable notifyInvalid, if true, it will send an editor warning that it found invalid tokens. This is defaulted to true so unless specified, the function will make an editor warning.

### Important Notes

     - SayDialogs will truncate text if the limit is reached. No other runtime changes; everything else is in the editor.
- My change adds tests or test scenes

### Other information
**Important:** because I did a silly, this PR includes changes from PR #934 which contains an optional boolean to Writer Audio. When merging this branch, 934 can be deleted as this already contains it. 

Future updates could include having specific say commands ignore the text limits, just in case your Say Dialog is using Best Fit and you want just a couple commands to be a little longer than the rest, if such a feature is not added prior to merging this PR.


CG-Tespy:
This is pretty great! ^_^

Arylos07:
> This is pretty great! ^_^

Thanks! It's not the prettiest backend solution, but it was the best way I could get it to work but has been very helpful.

breadnone:
Adding text elipsis as an option to this would be nice. So when the text gets truncated, it will show `...` (triple dots or something like that)

Arylos07:
> Adding elipsis as an option to this would be nice. So when the text gets truncated, it will show `...` (triple dots or user customizable), indicating that there's a continuation to the sentence. Just an idea, otherwise feel free to ignore it :)

The built in Truncate function added that the say dialog uses does have "..." by default, so any time you call it without supplying a different suffix, it will use that instead. The reason it does this is because the truncated text doesn't actually show up in a different dialog box or anything; it serves as a warning to the developer from the game that the text is too long, as a way to debug just in case the counter was mixed. This can be changed in the Say command by changing what suffix it uses.

TheEmbracedOne:
Thank you for this, this is exactly what I needed!

Issue #910, opened 2020-10-25
Say command's text skips ahead, based on previous Say command's text length


TheEmbracedOne:
**Describe the bug**
My Say dialogue's Writer script is set to NOT write whole words, NOT instant complete, NOT do read ahead text, while writing speed was set to 30, yet, when the next Say command's text is displayed, it skips ahead. See gifs below. 

**To Reproduce**
1. Create a scene with the following setup: 
**Say dialogue**
fade duration: 1

**Writer**
writing speed: 30
write whole words: off
instant complete: off
do read ahead text: off

Make sure both Say commands are set to "Always Show" and untick "Fade When Done". Menu commands should NOT be "Hide if Visited".

**Commands**
- First node to have a Say command and a Menu command that leads to the second node
- Second node to have a Say command and a Menu command that leads back to the previous node. 

2. Enter play mode, click the menu option, observe the text being written out and remaining onscreen.
3. Click the menu option to go back to the previous node.
4. Click the menu option again, observe text not being written out, but instead some of it appearing onscreen, as if the previous Say command's text's length affected it. See gifs below.

**Expected behavior**
Text in Say commands set to "Always Show" to be written out from scratch every time they're displayed.

**Screenshots**
Say text skipping ahead based on previous Say text's length:
![ePmo1eXO0P](https://user-images.githubusercontent.com/16197906/97114063-2db41d80-16e6-11eb-882d-86cd292a0c61.gif)

Same thing, but I made the first Say text 1 character. Observe the difference:
![grkK4dggon](https://user-images.githubusercontent.com/16197906/97114303-98b22400-16e7-11eb-886b-022f1f0a7a09.gif)

Here, you can see the same text being displayed for the first time, being written out, then second time, it doesnt get written out, it just fully appears onscreen. It would be expected that it always gets written out from scratch like first time:
![TCe9kG1Vjm](https://user-images.githubusercontent.com/16197906/97113922-4a038a80-16e5-11eb-80b5-b78b5f1377fa.gif)

breadnone:
Confirmed, I've this issue with texts instantly revealed right after interacting with menu dialog

breadnone:
Man, this problem is all over my game now... Tried to track it down but found nothing

TheEmbracedOne:
Unfortunately the pr by @breadnone #952 hasnt fixed the issue - here's a slower gif to demonstrate:
[...]

PR #952, opened 2021-02-01
nextLineInputFlagWon't Always Be True On Mouse Clicks

breadnone:
### Description
Currently the DialogInput will always set nextlineinputflag to true when clicking mouse multiple times even if there's no writing activity on screen, thus making the next Say auto-completed. Even mouse clicks when there's menu choices will trigger the next Say to be auto-completed.

This may solves:
https://github.com/snozbot/fungus/issues/910
https://github.com/snozbot/fungus/issues/954

Every mouse clicks will set nextlineinputflag  to true

### What is the new behavior?
Add condition to SetNextLineFlag so it won't always be true on button presses when there's no writing activity on screen or when the text is already completed.

Before:  

https://user-images.githubusercontent.com/64100867/106449306-69653d00-64b6-11eb-8c3b-db07ff979a9c.mp4

After:

https://user-images.githubusercontent.com/64100867/106449364-7aae4980-64b6-11eb-912e-bfde64758b51.mp4

### Other information

Previous, this auto-complete bug is all over my game. It also gets triggered when there's menu choices. So far, I've not encountered that glaring issue anymore with this patch.

Note: It works for my game, but still needs further testing!

Edit: I could not reproduce the issue with text automatically completed after clicking the button on menu choices. Probably fixed?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,18 +8,18 @@ assignees: ''
 ---
 
 **Describe the bug**
-A clear and concise description of what the bug is. Is it a bug during playback or a bug in editing Flowcharts, Blocks, Commands, and so on? When and where it is happening, is it only in editor or only in build or both?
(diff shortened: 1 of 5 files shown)


breadnone:
Pardon the edits!

stevehalliwell:
Thanks for the PR.

PR #953, opened 2021-02-01
Fix GetSecondsRemaining

KVinS:
### Description
While doing lipsync, I ran into the fact that it is impossible to understand when the voice over clip finishes its playback.

After playing a voice, the GetSecondsRemaining returns its full duration audio clip.

### What is the new behavior?
After the voice clip has finished playing, the number of seconds remaining becomes 0.


stevehalliwell:
Thanks for the PR.

PR #955, opened 2021-02-01
Make nextClickDelay works with ClickAnywhere 

breadnone:
### Description
Similar to ClickOnDialog for dialog progression. It has capability to delay the next input, while ClickAnywhere does not

nextClickDelay does not work with ClickAnywhere

### What is the new behavior?
nextClickDelay now works with ClickAnywhere

### Important Notes
- My change does not require modifcations or additions to documentation


stevehalliwell:
Thanks for the PR.

PR #968, opened 2021-04-26
[Bugfix] Drag Event Handlers Executing for Draggables not Really Being Dragged

CG-Tespy:
### Description
Fixes a bug where drag events can be invoked even when a draggable isn't actually being dragged by the player

### How to Reproduce the Bug
Open the DragAndDrop example scene, and try moving the DraggableSprite using editor controls, instead of the usual way. Move it onto the DragTargetSprite, and then off of it; both DragEntered and DragExited will execute.

### What is the new behavior?
Thanks to a bool implemented in Draggable2D (and drag events altered to take it into account), DragEntered and DragExited only execute for draggables actually being dragged (be it by the player, or code meant to simulate the act).

### Important Notes
* Moved the Drag Event Handlers into their own folder called "Drag" for better organization
  *  Figured I might as well
- My change modifies the runtime execution/behaviour of Draggable2D, DragEntered, DragExited and DragCompleted

### Other information
* I discovered this bug while working on a game, where as the player moves a layout group of draggable objects, DragEntered and DragExited execute from being moved by the Scroll Rect


stevehalliwell:
Thanks for the PR. Looks like the history has some old commits in it so I'll squash it during the merge.

PR #969, opened 2021-05-02
Retain Command.ItemId on Cut and Paste

ongjinwen:
### Description

Do not assign new ItemId to pasted commands on Cut and Paste. This is causing issue for me as I use ItemId to manage saves in my project.

Paste always assign a new ItemId to pasted commands.

### What is the new behavior?

- Pasted commands will retain ItemId
- Cut and Paste will clear CopyCommandBuffer

### DIFF
--- a/Assets/Fungus/Scripts/Components/CommandCopyBuffer.cs
+++ b/Assets/Fungus/Scripts/Components/CommandCopyBuffer.cs
@@ -70,6 +70,8 @@ namespace Fungus
             }
         }
 
+        public bool IsCut { get; set; }
+
         #endregion
     }
 }
--- a/Assets/Fungus/Scripts/Editor/BlockEditor.cs
+++ b/Assets/Fungus/Scripts/Editor/BlockEditor.cs
@@ -656,6 +656,8 @@ namespace Fungus.EditorUtils
         {
             Copy();
             Delete();
+
+            CommandCopyBuffer.GetInstance().IsCut = true;
         }
 
         protected void Copy()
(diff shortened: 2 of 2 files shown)


stevehalliwell:
Can you clarify the use case here?

I'm also concerned that this will introduce duplicate commandIDs when cut and pasting commands occurs between 2 flowcharts. Which I think is probably, primary use case of cut and paste right now.

breadnone:
Perhaps reconsidering this pr before merging. Currently, Fungus would sometimes/reandomly having duplicates IDs when copied/pasted for if-else commands and Fungus will only consider the 1st copy. This may make it worse

ongjinwen:
> Can you clarify the use case here?
> 
> I'm also concerned that this will introduce duplicate commandIDs when cut and pasting commands occurs between 2 flowcharts. Which I think is probably, primary use case of cut and paste right now.

I see. I wasn't aware that commands could be copied between flowcharts, as I have only tested it within blocks of a flowchart. Let me revise my changes.

I wanted a save (and history) system that implicitly tracks commands instead of using save points. For this, I require a way to uniquely identify the command objects. I looked around the properties and it seems item ID fits my use case. However, the current cut-paste behaviour (basically a copy-delete-paste) always initialize a new ID to the new command. Oh, and cut-paste should erase the command buffer too. :P

> Perhaps reconsidering this pr before merging. Currently, Fungus would sometimes/randomly having duplicates IDs when copying/pasting if-else commands and Fungus will only consider the 1st copy and completely ignore the other. This may make it worse.
> 
> No bug report for this yet, bcos it's super random/rare and very hard to reproduce, but people on Fungus' discord server has been randomly encountered this issue every now and then

Noted, I see what I can find.

TheEmbracedOne:
@ongjinwen Dont know if this issue is still relevant but you could also use the blockname + command index combo to do what you described above. CommandIndex is the index of the command in its parent block (ie first command is 0, then the next is 1, the next is 2 and so on). This could probably solve the issue of pasting perhaps, unless you figured it out already.

Hope this is helpful.

PR #971, opened 2021-05-02
Fix double duplication of blocks with Ctrl+D

ongjinwen:
### Description

Fix bug where Ctrl+D will duplicate selected blocks twice.

1. Select blocks
2. Press Ctrl+D to duplicate
3. Every selected blocks will be duplicated twice (one of them hidden behind the other)

### What is the new behavior?

- Union blocks before copying


stevehalliwell:
Thanks for the PR.

PR #972, opened 2021-05-02
Fix visual bug when dragging blocks with FungusEditorPreferences.useGridSnap enabled

ongjinwen:
### Description

Minor refactoring to fix visual bug when FungusEditorPreferences.useGridSnap is enabled.

When FungusEditorPreferences.useGridSnap is enabled, labels and connections do not follow the snapped position of block._NodeRect. Apparently, this change also fixes the snap width when scrolling.

### What is the new behavior?

- Store and use temp block rect values
- Manage in OnMouse methods instead of Draw methods

### DIFF
--- a/Assets/Fungus/Scripts/Editor/EditorZoomArea.cs
+++ b/Assets/Fungus/Scripts/Editor/EditorZoomArea.cs
@@ -56,9 +56,8 @@ namespace Fungus.EditorUtils
         public static Rect SnapPosition(this Rect rect, float snapInterval)
         {
             var tmp = rect;
-            var x = tmp.position.x;
-            var y = tmp.position.y;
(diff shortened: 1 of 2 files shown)


ongjinwen:
FlowchartWindow.cs is quite messy... X_X

stevehalliwell:
Thanks for the PR.

Clarifying, the minor, purely visual bug mentioned is the connecting lines not matching perfectly the snap point during while user is dragging.

Having looked at the severity of the issue and the changes to the code, I'm going to remove this from the current bugfix. I'll return to it soon. At that time I'll most likely merge it in and then make changes to the solution here to remove the need for the dictonary, as it makes the source of truth for block positions unclear and makes future work more error prone.

Issue #961, opened 2021-03-29
JSONChecker.cs uses deprecated code.


ianicmathieu:
**Describe the bug**
Assets\Fungus\Thirdparty\FungusLua\Thirdparty\JSON\Editor\JSONChecker.cs(76,28): warning CS0618: 'UnityWebRequest.isNetworkError' is obsolete: 'UnityWebRequest.isNetworkError is deprecated. Use (UnityWebRequest.result == UnityWebRequest.Result.ConnectionError) instead.'

Not exactly sure why this is popping up in 2020 LTS but in my project, I have this warning.

Looks like some code needs to be updated.

ianicmathieu:
while (!test.isDone && test.result == UnityWebRequest.Result.ConnectionError);
(edited)

The fix appear to be change the unitywebrequest line to this.

stevehalliwell:
Just from the file path, seems to come from a thirdparty dep. So we'll either check there for updates or consider bringing it out  into our own fork, if the license allows.

PR #986, opened 2021-06-13
Update deprecated isNetworkError

breadnone:
### Description
replace deprecated isNetworkError with UnityWebRequest.Result.ConnectionError

### Other information

closes https://github.com/snozbot/fungus/issues/961


scotthammack (changes requested):
Shouldn't this be a != since the original was checking for !test.isNetworkError?

Issue #973, opened 2021-05-03
Skipping flashing dialogue causes stackoverflow exception.


puetsua:
**Describe the bug**
Skipping flashing dialogue causes stackoverflow exception.

**To Reproduce**
1. Go to the scene 'NarrativeLog' in the FungusExample
2. Add `{flash=0.5}` in front of every Say commands.
3. Hit Play
4. Keep pressing space or pressing escape to **skip** a dialogue during a flash. (Almost always reproduce.)

**Expected behavior**
No exception when skipping a dialogue with flashing effect.

**Screenshots**
![image](https://user-images.githubusercontent.com/9292138/116835031-43ede980-abf3-11eb-9e73-f3983f528cd7.png)
![image](https://user-images.githubusercontent.com/9292138/116835153-d7bfb580-abf3-11eb-98cf-e6bb6f7e24d4.png)

**Versions & Platform (please complete the following information):**
 - Unity version 2019.4.24f1
 - Fungus Version v.3.13.4
 - Editor OS: Win 10
 - Build Platform //

breadnone:
I'm guessing, this happens because the flash command itself doesn't wait from one to another. In other words, when you're cancelling them while the previous `flash` still flashing then this would happen.

If my memory serves me right, cancelling dialogues (Esc), ignores `WaitUntilFinished`. To disable `Cancellling with Esc button` you can toggle off `Cancel Enable` on your SayDialog component OR just add a longer input delay per/second from your EventSystem in the hierarchy.

puetsua:
> 
> 
> I'm guessing, this happens because the flash command itself doesn't wait from one to another. In other words, when you're cancelling them while the previous `flash` still flashing then this would happen.
> 
> If my memory serves me right, cancelling dialogues (Esc), ignores `WaitUntilFinished`. To disable `Cancellling with Esc button` you can toggle off `Cancel Enable` on your SayDialog component OR just add a longer input delay per/second from your StandAloneInput module.

You're right! I tried to set flash time to 5 seconds and it triggered the bug if I put another flash while the previous hasn't finished yet. 

I don't think I will disable `Cancel Enable` because it comes in handy when I'm testing my dialogues. I probably will try to fix it myself first. Maybe just cancelling previous the flash and run next one.

puetsua:
Not sure if this is elegant way to fix it but at least it works now.

CameraManager.cs:
```c#
namespace Fungus
{
    public class CameraManager : MonoBehaviour
    {
        // ...

[...]

PR #988, opened 2021-06-13
Fix StackOverflow Exception When Skipping While Flash Still Active

breadnone:
### Description
When skipping dialogue while Flash still not done tweening, it will cause StackOverflow Exception

Skipping dialogue will throw StackOverflow Exception while flash is not done tweening

### What is the new behavior?
Will not re-instantiate tween if Flash is still active 

### Other information

Closes https://github.com/snozbot/fungus/issues/973


breadnone:
This fix doesn't make any sense... It would just cause problems to anything related to stopping the fading prematurely.

Just don't set Cancel to enable. That feature doesn't make any sense on build

Closing this...

Issue #945, opened 2021-01-18
Narrative Log needs a character count to prevent it from breaking all UI text in games.


ianicmathieu:
**Describe the bug**
If you have a lot of text in the Narrative log, Unity will hit it's vertices limit and all subsequent text fields will show garbage text or be empty

ArgumentException: Mesh can not have more than 65000 vertices

**To Reproduce**
Simply fill up the Narrative log with text until unity starts giving the error.
At some point - when 65000 vertices are reached, your text will become garbage in the narrative log and everywhere else.
This pretty much breaks your game.

A solution could be to control the character count of the Narrative log.

PR #989, opened 2021-06-14
Limit Characters Shown In NarrativeLog to Avoid 65000 Vertices Limit

breadnone:
### Description
Fixes 65000 vertices limit by limiting character shown in NarrativeLog

It will throw **ArgumentException: Mesh can not have more than 65000 vertices** if NarrativeLog texts over 65000 vertices

### What is the new behavior?
The text will get truncated to default 10000 characters (users can change the characters limit) and a tripple-dots prefix will be added as indicator that it was truncated

**BEFORE**  

https://user-images.githubusercontent.com/64100867/121877827-3faa0580-cd35-11eb-8fba-3f2e83578e8c.mp4

**AFTER**  

https://user-images.githubusercontent.com/64100867/121878023-708a3a80-cd35-11eb-9a30-881bf98f9f23.mp4

### Other information

Edit: Description edit

Relates to https://github.com/snozbot/fungus/issues/945

### DIFF
--- a/Assets/Fungus/Scripts/Components/NarrativeLog.cs
+++ b/Assets/Fungus/Scripts/Components/NarrativeLog.cs
@@ -35,6 +35,7 @@ namespace Fungus
         /// </summary>
         public static event NarrativeAddedHandler OnNarrativeAdded;
         public delegate void NarrativeAddedHandler(NarrativeLogEntry data);
+        public static int maxSize {get; set;} = 10000;
         public static void DoNarrativeAdded(NarrativeLogEntry data)
         {
             if (OnNarrativeAdded != null)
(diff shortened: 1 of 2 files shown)


breadnone:
Stumbled upon this issue https://github.com/snozbot/fungus/issues/945 when debugging my game. I'm not sure if this is the correct implementation @stevehalliwell 

Feel free to close it if it's not desired. Thanks!

breadnone:
Another solution as describe here https://docs.unity3d.com/ScriptReference/Mesh-indexFormat.html

stevehalliwell:
Haven't double checked but I don't think we can change the internal mesh format used by the text elements.

Using the alternate entry based Narative Log, submitted by CG-Tespy, that was added a while back would also avoid the issue.

stevehalliwell:
I think this is a good quick fix . Proper pagenation being a better long term goal.

I'm going to manually merge and make some changes though so the implementation is entirely within the narrativelog view code though.

PR #996, opened 2021-06-19
QoL Update: Select and scroll to new command on add

### LINKED ISSUE #994: Newly Added Command Won't Auto Focused/Selected After Creation
**Describe the bug**
Newly added commands would not be focused after it gets created thus making it hard to search for it if we have many commands in that same block.

Another weird behavior, is that after the we clicked one of the command in inspector, then add new command, the newly added command would be placed under it (this is an expected behavior, from code) BUT if we add other commands after that, they would be placed at the most bottom order in that block which kinda weird.

**To Reproduce**
Add any new commands, and once added, the block inspector would be blank or out of focus.

**Expected behavior**
Should be automatically auto focus to the newly added commands

**Versions & Platform (please complete the following information):**
 - Unity version 2020 LTS

Arylos07:
### Description
Minor quality of life update. When a new command is added from the command list, it will automatically select that new command and also scroll to it in the inspector if needed. Subsequently added commands will also go underneath the new command instead of going straight to the bottom.

Note: This does not affect copy + paste or duplicate. That can be addressed in a different PR since that's kind of niche. This only affects adding new commands from the commands menu.

Additionally, this is untested with the old menu layout. It *should* work though. If not, I'll add a toggle so this feature is disabled when using the old ui.

Currently, when a new command is added, it goes to whatever the currently selected command is, which isn't bad, but it deselects commands, requiring you to click on the new command in order to modify it.

Additionally, if adding commands in sequence (repeatedly adding them), any additional commands will go to the bottom of the list. This isn't bad unless you're trying to set up stuff in the middle of an already configured block. 

Resolves #994 

### What is the new behavior?

- The adding of commands is streamlined so when you add a command, it's already selected for editing and adding new commands under it.

### Important Notes

     - Only partially. Those who are used to clicking new commands will be affected.

### Other information
**Important:** because I did a silly, this PR includes changes from PR #934 which contains an optional boolean to Writer Audio. When merging this branch, 934 can be deleted as this already contains it. (as usual)

### DIFF
--- a/Assets/Fungus/Scripts/Components/WriterAudio.cs
+++ b/Assets/Fungus/Scripts/Components/WriterAudio.cs
@@ -45,6 +45,9 @@ namespace Fungus
         [Tooltip("Sound effect to play on user input (e.g. a click)")]
         [SerializeField] protected AudioClip inputSound;
 
+        [Tooltip("Lerp voice effect while playing. Enabled results in smoother volume transitions between different write sounds. Disable to keep writer volume constant.")]
+        [SerializeField] protected bool lerpWriterVolume = true;
+
         protected float targetVolume = 0f;
 
         // When true, a beep will be played on every written character glyph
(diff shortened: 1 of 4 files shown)


stevehalliwell:
Thanks for the PR. 

After reviewing it I'll manually checkout and cherrypick as it seems you've got some unreleated commits here.

breadnone:
Tested this, since yesterday, didn't find any issue

stevehalliwell:
Cherry-picked commits directly related to this issue from the submitted branch.

Issue #992, opened 2021-06-16
Feature Request: Lock Flowchart


TheEmbracedOne:
**Is your feature request related to a problem? Please describe.**
I have a prefab that houses all of my fungus components that is present in all my scenes. When I need to modify a flowchart that is part of this prefab and I enter prefab mode, but if I do that while having selected the flowchart, I need to click the flowchart again while in prefab mode to edit the prefab version of it. 
Sometimes it happens that I forget this and I end up editing a flowchart accidentally NOT in prefab mode, leading to a lot of cherrypicking stuff to revert/apply to the prefab.

**Describe the solution you'd like**
I'd like there to be a way to lock a flowchart from being edited until the lock is removed; no commands may be modified, no commands may be removed or added, or blocks moved.

This way I could make sure I only edit my flowcharts in prefab mode, but I think the use for this feature is wider range than that, a user may want to lock a flowchart they're finished working on and so on.

breadnone:
This should solve your problem https://docs.unity3d.com/ScriptReference/SelectionBaseAttribute.html

TheEmbracedOne:
but isnt the SelectionBaseAttribute to do with scene view selection?

> In the Unity Scene View, when clicking to select objects, Unity will try to figure out the best object to select for you.

I'm not selecting anything in scene view. All I'd like to do is lock the flowchart and all of the scripts on it from editing, so no commands or blocks can be added/removed/edited.

Arylos07:
I think I may have a solution for this. I'll do some research over the weekend and report back.

TheEmbracedOne:
@Arylos07 thanks, I look forward to hearing about your findings!

Arylos07:
Okay, I definitely have a solution that is pretty lightweight. I'll try and get that added by Monday as a separate pull request.

Arylos07:
#997 Adds this ability now. I feel like it can be a bit jank, but it is the best implementation I can think of without doing anything memory intensive like caching values.

Let me know what you think.

TheEmbracedOne:
Cheers @Arylos07 i've added this now, I'll report back if I run into any issues!

TheEmbracedOne:
This has been working wonderfully for the past weeks, I'll close the issue for now.

TheEmbracedOne:
Hi @Arylos07 I lef this message on Discord as well, but I'm leaving it here for transparency:

[...]

PR #997, opened 2021-06-20
New Feature: Lockable flowchart

Arylos07:
### Description
Added an update to the flowchart window and editors so flowcharts can be locked, preventing editing. 

Currently, when working with flowchart prefabs and defaults, you may want the flowchart to not be editable at least not without needing to. This can also be useful for tracking what flowcharts are done and not. 

Issue Number: #992 

### What is the new behavior?
Clicking on the padlock icon in the flowchart window (or "lock flowchart" in the flowchart inspector) will lock the flowchart. It will not keep the flowchart in the editor window, unlike locking an inspector, but will prevent the blocks, commands, metadata, and variables from being changed. However, you can still scroll through commands to view them and move through the flowchart freely. It does work by greying out the properties, however. Will edit later if needed.

ezgif-3-5d24a5c22987

In the future, the flowchart window can be adjusted to also lock on one flowchart, not changing if another is selected. However this will only be useful when multiple flowchart windows can be opened at once; a feature that seems to be missing for now.

### Important Notes

     - Possibly. Flowcharts are unlocked by default so it shouldn't change much.

So far, no breaking changes have been found and it should also merge with #996 with how I edited the code. One main issue is that users may see the padlock and think it locks the flowchart in that window when it does not; only locks editing of it. This will be addressed when multiple flowchart windows can be open, at which point an additional behaviour will be added.

### Other information

**Important:** because I did a silly, this PR includes changes from PR #934 which contains an optional boolean to Writer Audio. When merging this branch, 934 can be deleted as this already contains it. (as usual)

### DIFF
--- a/Assets/Fungus/Scripts/Components/Flowchart.cs
+++ b/Assets/Fungus/Scripts/Components/Flowchart.cs
@@ -24,6 +24,9 @@ namespace Fungus
         [SerializeField] protected int version = 0; // Default to 0 to always trigger an update for older versions of Fungus.
 
         [HideInInspector]
+        [SerializeField] public bool locked = false;
+
+        [HideInInspector]
         [SerializeField] protected Vector2 scrollPos;
 
         [HideInInspector]
--- a/Assets/Fungus/Scripts/Components/WriterAudio.cs
+++ b/Assets/Fungus/Scripts/Components/WriterAudio.cs
@@ -45,6 +45,9 @@ namespace Fungus
         [Tooltip("Sound effect to play on user input (e.g. a click)")]
         [SerializeField] protected AudioClip inputSound;
 
+        [Tooltip("Lerp voice effect while playing. Enabled results in smoother volume transitions between different write sounds. Disable to keep writer volume constant.")]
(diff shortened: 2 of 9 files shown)


Arylos07:
Resolved a merge conflict with CommandListAdaptor after my last changes were merged. I saw this coming so it was simple.

TheEmbracedOne:
Hi @Arylos07 I lef this message on Discord as well, but I'm leaving it here for transparency:

I've been using your lock flowchart feature ever since you PR'ed in, and I only now ran into a bug with it, where it prevents editing the flowchart itself, I mean the actual Flowchart.cs on the gameobject, it seems like it redraws itself constantly if I try to edit it EVEN if the lock is off. I went back to a version right before I added the PR and the bug is not present there. Could you please take a look at this issue sometime? Thank you

DaIBQH5dSz


Gif added to demonstrate (as soon as I click an editable field or open the inspector window selection thing, it gets redrawn)

TheEmbracedOne:
Can confirm the fix submitted fixes my issue! Thank you again @Arylos07

PR #1014, opened 2021-07-10
[Incomplete] Refactoring Save Slot Controllers to work more with Save Slot Views

CG-Tespy:
### Description
It's a refactor.

I've noticed that the source for the new SaveSlotController class has yet to be fully adapted to the new SlotView system. I figured I might as well take a crack at making it so, even writing unit tests to make sure things are as intended. At the time of this writing, I've only managed to confirm that the slots do indeed have their base views registered as intended.

Later on, I'd like to make changes such as removing the old nameText, descText, and timeStampText fields; the new view system renders them unnecessary.

I hope to later on also help clean up other UI-related parts of the save system, though that's for another time.

### Important Notes
- At the time of this writing, there aren't any real behavior changes
- My changes require modifications or additions to documentation
- Added a GetView function to SaveSlotController so that other objects can access particular views of any particular slot. Certainly helped with unit-testing.
- Please see the changes to SaveSlotController's Start function
- My change adds unit tests under the Tests folder. 
-  I've also added a Resources folder to help manage testing-only resources such as the

### DIFF
--- a/Assets/Fungus/Resources/Prefabs/SaveSlot.prefab
+++ b/Assets/Fungus/Resources/Prefabs/SaveSlot.prefab
@@ -128,7 +128,7 @@ GameObject:
   - component: {fileID: 8335079148700219652}
   - component: {fileID: 4849907420850878217}
   m_Layer: 5
-  m_Name: Name
+  m_Name: Number
   m_TagString: Untagged
   m_Icon: {fileID: 0}
   m_NavMeshLayer: 0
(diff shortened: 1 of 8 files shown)


CG-Tespy:
@stevehalliwell Finished. Removed some legacy fields and added tests to ensure that the slots do their thing right on their own. I've added to the view classes as well, based on the things the unit tests needed to check for.

PR #1017, opened 2021-07-15
[Feature] Basics of Save Categories

CG-Tespy:
### Important Notes
* Introduces basic aspects of a save category system. 
  * The idea is to make it easier for the user to have different types of save files in their game, like how Fire Emblem has normal saves, autosaves and battle saves. 
* This is also to help make the system more flowchart-friendly. Future PRs may involve using the stuff from this one to help refactor and clean up the current system's codebase
* Did a little reorganizing in the Prefabs folder so that the save-system-related prefabs are in their own subfolder


stevehalliwell:
A similar concept exists in the consolidate and expand draft PR work,

CG-Tespy:
> A similar concept exists in the consolidate and expand draft PR work,

Which point in that PR is it? I read through each, and none seem all that similar to what I'm proposing here

Issue #627, opened 2017-08-12
Timeline / Playable -- Start/Stop actions for Unity 2017.1


astraldata:
The new Cinemachine / Timeline integration in 2017.1 is so great with all the storytelling stuff it lets you do, but as soon as I saw it, I totally thought about Fungus and thought "this would be more amazing if we could pause and resume a timeline after a few blocks of dialogue, play some animation, then have Timeline pause itself to resume Fungus dialogue until the end of a block, then letting the Timeline finish when the block is done!

This would essentially let Fungus piggyback off of Unity's advanced animation sequencer while running its own dialogue starts and stops when it needs to start, resume, or stop a block in Fungus to let a Timeline play. Unity's new "Playables" feature is right up the alley of Fungus's mission, IMO, since it requires no code to use and its purpose is nothing more than to tell a story!

chrisgregan:
My job is eating all my dev time these days but yeah a Timeline integration would be a great addition. If anyone has time to put together some commands and example scenes for it I'd be happy to review a pull request.

stevehalliwell:
I'd be keen to. But won't be for a few weeks as we've not moved to 2017 at work yet.

sinistersnare:
Is Fungus usable with Timeline/Cinemachine? Is there documentation on interactions between the two?

PR #1024, opened 2021-08-18
Added PlayTimeline command

GuroGuru:
### Description
This PR is a new feature, the command PlayTimeline.

Currently Fungus animation-related commands don't compreehend the Unity Timeline functionality, importing sequencing tool.
Issue Number: #627 

### What is the new behavior?
The new command PlayTimeline allows specifying a Timeline object from the scene and play it during the flow. It also allows choosing whether if the flow will be blocked until the timeline stops or if it will continue the execution.

### Important Notes
- My change adds tests or test scenes

### DIFF
--- a/Assets/Fungus/EditorResources/FungusEditorResources.asset.meta
+++ b/Assets/Fungus/EditorResources/FungusEditorResources.asset.meta
@@ -1,8 +1,7 @@
 fileFormatVersion: 2
 guid: 29cae11326cb84802b83fc3ff25a47e8
-timeCreated: 1495441032
-licenseType: Free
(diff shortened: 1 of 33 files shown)


breadnone (changes requested):
Perhaps restructure it better to be inline with Fungus coding standards. Any wait operations usually called waitUntilFinished.

And you can wrap your if condition for continue inside your delegate function to reduce code duplication

Something like this
``` 

            timeline.stopped += delegate 
            {
               if(waitUntilFinished)
               Continue();};
            }
            if(!waitUntilFinished)
            {
                Continue();
            }```

Issue #921, opened 2020-11-21
"Hide if Visited" and "Show Count" not "remembered" after loading save


TheEmbracedOne:
**Describe the bug**
If a menu option is marked "Hide if Visited", it is not "remembered" by Fungus if the player quits the game and reloads.

**To Reproduce**

1. Have a for "main menu" where you can click a Load button that just executes SaveMenu.Load(). The SaveMenu prefab exists here and will persist through to the next scene. SaveMenu has load on start OFF, auto save ON

2. Have a "test scene" where selecting "new game" or "load" takes you. It should have the following flowchart setup:
**Start Block** with a menu option that's marked "Hide if Visited", leading to "Block2". 
"**Block2**" should have a SavePoint, then a Say dialogue text, then a menu option that leads back to the Start block, not marked "Hide if Visited".

### Test Scenarios
Test scenario 1: we click the menu option to go to "Block2". we see the dialogue text come up, then the "back" option. We quit the game without clicking it. Upon relaunching the game, we are put on Block2, we see the say text, then the "back" option. We click "back", and we find ourselves on the Start block to observe the option that would lead to Block2 gone.

Test scenario 2: we click the menu option to go to "Block2". we see the dialogue text come up, then the "back" option. We click the "back" option to get back to the Start block and observe that the option leading to "Block2" is gone. We quit the game now. Upon relaunching the game, we are put on the Start block, but we observe that the option taking us to "Block2" is available once again.

### "Show Always"/"Show Count" Interactions
If "Show Always" is unticked, I can specify how many times a say dialogue appears via "show count". if set to appear only once, then in Test Scenario 2 if I set the say dialogue to have "show always" OFF and "show count" set to 1, the say dialogue will appear upon loading the save.

**Expected behavior**
"Hide if Visited" to persist after reloading or relaunching the game.
"Show Count" to persist after reloading or relaunching the game.

**Screenshots**
![image](https://user-images.githubusercontent.com/16197906/99884917-57029380-2c29-11eb-8eab-e98d2f44ee57.png)

![image](https://user-images.githubusercontent.com/16197906/99884584-4d782c00-2c27-11eb-814a-e030c778a3c0.png)
![image](https://user-images.githubusercontent.com/16197906/99884892-3fc3a600-2c29-11eb-9085-16e49631c31d.png)
![image](https://user-images.githubusercontent.com/16197906/99884899-46521d80-2c29-11eb-9f9c-4cdee058fbaa.png)

**Versions & Platform:**
[...]

PR #1026, opened 2021-08-25
[WIP] Unit tests for creating save data

CG-Tespy:
### Description
I felt we need to solidify more of the more basic stuff, such as creating save data of specific types, starting with the numeric Fungus Variables that Flowcharts can have.

### Important Notes
* I have the unit tests work by loading prefab scenes placed somewhere in the main Test folder. Said tests then using the prefabs' contents to do their thing. It's a pattern I've found more useful (and more fun) than having to set up scenes entirely in code.
* I felt it'd be good to have the encoders be MonoBehaviors so that users can swap them in and out as they see fit, hence what you see attached to the FlowchartEncoders GameObject
  * After all, what if the user wants to change how the built-in data types are saved? I believe it's best to give them that option by
  * This would also mean that the main FlowchartEncoder (not yet coded in) would have to be told which GameObjects contain its subencoders, allowing for more organization
* My change adds tests or test scenes
  * Naturally
* From what I've gathered from reverse-engineering this system, StringPairs are intended to be the main containers of off-disk saves. Hence why the encoder returns instances of those
* For the sake of the SRP, I felt it's be best to have various types have their own encoders, hence the NumericVarSaveEncoder class for the numeric Fungus Vars
* Before becoming an official collaborator on Fungus, I did a little work on saving sound-related data. You can ignore that stuff involved in this PR; it'll likely get replaced by other code later down the line

### Main To-Do List
* Saving the following essential var types: (done!)
  * Numerics (ints, floats, bools)
  * String
  * Color
  * Vectors (2D, 3D, 4D)
  * Transform
* Saving the state of Fungus's Portrait system (done!)
  * What portraits were shown at the time the game was saved
* Refine the implementation as appropriate
* Saving the execution states of Flowcharts (done!)
  * The execution counts of the blocks
  * The execution counts of Say Command
* Keeping the states of the Menu blocks. Problem mentioned in #921 (done!)
  * Hide If Visited
  * Interactable
  * Hide This Option 

### Secondary To-Do List
* Saving the following (not quite as essential) var types:
  * Rigidbodies (2D, 3D)
  * Sound Clip
  * Audio Source
  * Quaternion
* Saving the state of the Fungus's Audio system
  * What BGM and Ambient sounds were playing
  * **May need to wait on further implementation before starting on this**
* Saving the states of the following commands
  * Menu (done!)
  * Say (done!)
    * Has a counter for how many times it should show
  * Various Tweens
 * Saving the state of the current view


CG-Tespy:
What do you think, @stevehalliwell ?

CG-Tespy:
Added a to-do list

CG-Tespy:
At the time of this writing, we have things set up to save all the essential var types.

CG-Tespy:
While there's still stuff on the secondary to-do list, I think I've done enough work here to move onto another area of the save system for now

PR #1040, opened 2021-11-18
Error Checking and Detecting Unused Variables Functionality

TheEmbracedOne:
### Description
Added an error checking button and a checking for unused variables button to both the flowchart window and the flowchart editor. This serves as a "sanity check" that finds the issues in the current flowchart, colouring commands/blocks red, and/or prints out the issues found in the log so users may correct these problems faster.

Currently, if there is a missing targetBlock in a Call command, a missing returnValue in an IF command, it's difficult to find the problem. 

### What is the new behavior?
1. Pressing the "Check for Errors" button colours any command that contains an error in its summary, and any block that contains any command that contains error in its summary.
2. Pressing the "Check for Unused Variables" button prints out all unused variables.
      2.a If "remove unused variables" is enabled, it removes all unused variables from the list as well.

### Important Notes
- My change **may** require modifcations or additions to documentation (though maybe it's self-explanatory enough?)

---------------

Most commands already have "Error: [...]" in their summary when a targetBlock or parameters are null. Additionally:
* The GetSummary() of the Call.cs should now read "Error: targetBlock null!" when its targetBlock is null
* The GetSummary() of the Say.cs should now read "Error: character null!" when its character is null

(There are potentially other areas where the summary could be updated in this manner, for example menuDialog being null in Menu commands, MenuTimer having 0 time, VO clip missing from a Say command (probably needs a bool to enable this check, I presume some people dont plan to have VO) and so on.)

Adjusted the CommandListAdaptor, so that it colours the commands red when they contain an error, and their text white for readability, so users can find issues easier.

Added an override void "ErrorCheck" to the Command.cs, triggered by pressing the "Check For Errors" button in the flowchart editor or window. It prints out warnings in the log if it detects an IF or ELSEIF command having a null returnValue.

The IF and ELSEIF commands now can return their returnValue keys via GetReturnValueName() to the Flowchart.cs upon pressing the "Check for Unused Variables" button in the flowchart editor or window, which compares the full list of variables against the list of returned, used variables.

### Screenshots
QNUIsvdY1u

hg5Bq3xDy1

image


TheEmbracedOne:
Invoke Method return values are now considered when checking for unused variables.

breadnone:
It should be hidden or at least two-level menu and confirmation whether it should be deleted or not

PR #1042, opened 2021-12-10
Unity Localization Package Support

addie-lombardo:
￼CleanShot 2021-12-09 at 09 16 50@2x

### Description
Hello! I recently released a commercial game with Fungus doing a lot of the heavy lifting including localization for dialogue. While this feature works as expected it proved to be difficult with larger projects so I wanted to create something that could be used from in small to large projects now that I have some free time. With the release of Unity 2021.2 and the public v1.0 release of the Unity Localization package I decided to reimplement Fungus's localization features to use the Localization package without removing the previous functionality in the codebase.

The current localization workflow requires you to write all your dialogue first, export it using the Localization component and then reimport it after editing the CSV in an external program.

### What is the new behavior?
The new localization workflow allows you to localize your content in two ways. 

The first way is the configure the LocalizedString variable in any ILocalizable component (named localized%VARIABLENAME% eg. localizedStoryText, localizedText, etc.) this is great for dialog that is repeated in multiple say commands and you’d like to assign them all to the same localization entry. 
￼
CleanShot 2021-12-09 at 12 30 37@2x

The second way is to automate the creation of entries and assigning of the LocalizedString variable using the new Localization component. The component doesn’t only allow you to export entries you can also import changes made to the string table back into fungus commands as well! Once exported to a string table you will be able to export to a variety of formats (CSV, Google Sheets, or XILFF) for external editing, all of which can be imported back in.
￼
CleanShot 2021-12-09 at 12 36 57@2x

PR also includes the ability to localize to now localize VO clips in the Say command thanks to the support to be able to localize assets with the Localization package.
￼
CleanShot 2021-12-09 at 18 10 58 2@2x

The new behavior includes two examples, a recreation of the original Localization example and a longer example which pulls dialogue from our game No Longer Home.

CleanShot 2021-12-09 at 19 48 45@2x

 The old localization system can still be used, all you need to do is remove UNITY_LOCALIZATION under Custom Defines in the Player Settings to revert back to using the old system.

### Important Notes

### Other information

I recommend [this tutorial](https://phrase.com/blog/posts/localizing-unity-games-official-localization-package/) for learning more about the basics and getting the Localization package setup.

Requires Unity 2021.2.0f1 to see the Localization package in the Package Manager but it can be installed in earlier editors by typing in com.unity.localization under “Add package from git URL” in the Package Manager in 2019.3 and later.

~~One feature I have not been able to reimplement yet (making the PR still a WIP) is Localization.SubstituteStrings() I would appreciate some help on context on what it does so I can implement it myself or get help.~~ Edit: I have implemented this in commit https://github.com/snozbot/fungus/pull/1042/commits/0d04f4775fde97001792d224a0cf403b219c42f3. I've updated to title to remove [WIP] as it is now feature complete


StefanoCecere:
I would love to see this PR merged!

PR #1045, opened 2021-12-31
[Feature] Timer System WIP

CG-Tespy:
### Description
This adds an easy-to-use Timer system to Fungus

### Why would Fungus need this?
It's pretty common to set time limits in games or check how long things took. It's even more common to track the player's playtime, yet as all official releases of Fungus (at the time of this writing) lack that, that forces devs to look to other things to implement those features.

With this, we save devs the trouble. **Plus, the new save system really needs this one;** it has to be able to show playtime in the UI, after all. Heck, that's the reason I even went through all this effort xD

### Important Notes
- My change modifies FungusManager
  - I needed a way to make sure that the TimerManager is in all scenes, and as I want this to be a built-in feature...
- My change adds unit tests
  - In the TimerSystem folder
- I have a tutorial right here: https://docs.google.com/document/d/1o5XB6SLRMc5HJ2v9lCYS_SCkeoGBLEnISQ7BRp9WIHw/edit?usp=sharing

### Other information
While I think the system is production-ready as-is, I would very much like to get feedback on it before it gets approved for an official release.


CG-Tespy:
I think things are looking pretty good now ^^ Thoughts, @chrisgregan and @stevehalliwell ?

PR #1046, opened 2022-01-03
Auto skip

addie-lombardo:
### Description
A common quality of life feature in visual novels is the ability to have the dialog continue after a period of time. I've implemented this feature into Fungus.

Currently you are required to click/press a button each time you want to advance the text.

### What is the new behavior?
Pressing the Tab key now toggles auto skip on and off. If enabled when a dialog entry ends it will wait the amount of time specified by Writer.AutoSkipWaitLength before continuing to the next command. If auto skip gets enabled after dialog has already finished printing it counts it as an input and continues to the next command. Additionally I've added a new serialized variable in SayDialog named autoSkipGO which takes a GameObject which will be toggle on and off to indicate to the player that auto skip is on (see GIF).

### Other information
CleanShot 2022-01-03 at 09 36 25

### DIFF
--- a/Assets/Fungus/Resources/Prefabs/SayDialog.prefab
+++ b/Assets/Fungus/Resources/Prefabs/SayDialog.prefab
@@ -2,15 +2,16 @@
 %TAG !u! tag:unity3d.com,2011:
 --- !u!1 &148914
 GameObject:
-  m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
-  m_PrefabInternal: {fileID: 100100000}
-  serializedVersion: 4
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
   m_Component:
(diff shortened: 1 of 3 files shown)


breadnone:
Wouldn't it better if it's just a coroutine right after the typing is done? similar to `Wait For Click`, this way we don't need to check every frame on Update and definitely much more cleaner.

addie-lombardo:
> Wouldn't it better if it's just a coroutine right after the typing is done? similar to `Wait For Click`, this way we don't need to check every frame on Update and definitely much more cleaner.
> 
> Edit: You don't event need a coroutine for this to work, a regular method would work wonders too, unless you want to put some sort of delays at the end before skipping, then yeah, coroutine is the way to go.

Fair enough! The code is actually from my game where we do have some delays/coroutines to yield to at the end hence the coroutine but I can definitely change it.

breadnone:
oh I see now, the Update just for checking button presses

Arylos07:
I was working on a similar feature that worked in a similar way, but allowed voice over correction and click-through. However, life got in the way before I could finish it. When my schedule improves, I'll look into opening a PR and adding it; it was nearly done but needed some thorough testing.

PR #1049, opened 2022-02-16
On UI Button Hover New Command

breadnone:
### Description
UI button mouse cursor state change. Some users asked this feature on discord to easily change cursor icon when hover on to UI button.

We don't have this command yet.

### What is the new behavior?
- When mouse hovered a button, the cursor icon/texture will change based on the assigned texture2d
- User can assign different Texture2d as their mouse cursor to different buttons (multiple cursor).
- Standard cursor mode
- Custom default cursor slot provided for when the game already uses custom mouse texture for fallback

Example video:

https://user-images.githubusercontent.com/64100867/154214872-7c5dc16a-41dd-475e-aaa8-6309eb67ce00.mp4

### Important Notes
- It compares GetInstanceID instead of object's name to avoid same name conflicts.

### Other information
- The new command will be in UI/OnMouseCursorHoverButton

### DIFF
--- a/Assets/Fungus/Scripts/Commands/UiButtonCursorState.cs
+++ b/Assets/Fungus/Scripts/Commands/UiButtonCursorState.cs
@@ -1,0 +1,125 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UI;
+using UnityEngine.EventSystems;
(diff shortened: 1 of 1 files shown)


breadnone:
I kinda think the AddListener isn't necessary. Just share your thoughts on this...

PR #1054, opened 2022-05-03
[WIP] Add toggle to run Fungus on unscaled time.

esklarski:
### Description
These changes enable Fungus to run while `Time.timeScale == 0`.

If `Time.timeScale` is set to `0` Fungus is paused and uninteractable

### What is the new behavior?
I made these changes, sans switch, for a project where I used Fungus for all the menus. I needed to pause 3d gameplay but have a character come up to act as the pause screen menu.

Behaviour Now:
- Fungus remains interactable while `Time.timeScale == 0`

### Important Notes
- My change adds demos to FungusExamples -- coming soon

### Other information
Still a work in progress:
- Need advising on location of behaviour switch (currently FungusManager.cs)
- I need to install the correct version of Unity before I can proceed
- Testing needs to happen
- I'll transfer the demo scene I put together

### DIFF
--- a/Assets/Fungus/Scripts/Commands/AudioMixerSnapshotTransitionTo.cs
+++ b/Assets/Fungus/Scripts/Commands/AudioMixerSnapshotTransitionTo.cs
@@ -27,7 +27,14 @@ namespace Fungus
 
             if (waitUntilFinished)
             {
-                StartCoroutine(WaitForTransition());
+                if (FungusManager.Instance.useUnscaledTime)
+                {
+                    StartCoroutine(WaitForRealtimeTransition());
+                }
+                else
+                {
+                    StartCoroutine(WaitForTransition());
+                }
             }
             else
             {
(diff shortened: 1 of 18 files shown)


breadnone:
Proly wrap it as extension instead of hard coding it?

esklarski:
I'm not sure extension methods can do all that this needs.

I'm more wondering, if there is any reason to pause Fungus? The simplest would be to eliminate all of the if/else logics and just run Fungus on unscaled time always.

breadnone:
That would introduce breaking change, lots of users pausing their games with it

esklarski:
Well that's why the effort. I hard coded it for my purposes, and didn't need to make a toggle.

esklarski:
Demo scene added.

esklarski:
I've done a bit more research and I can clean this up substantively.

Putting a YieldInstruction function like:
`public static YieldInstruction WaitForAdjustedSeconds(float seconds) =>
            useUnscaledTime ? new WaitForSecondsRealtime(seconds) : new WaitForSeconds(seconds);`
allows all of the if/else logic to be removed, with scripts only needing to call this one function instead of having a second wait function each and supporting logic:
`yield return FungusManager.Instance.WaitForAdjustedSeconds(value);`

It'll be a few days before I can cleanup the commit tree here.

PR #1064, opened 2023-05-14
New Feature: Listenable Say Dialogs

CG-Tespy:
### Description
As requested by someone on the Fungus Community Discord, I made a system that makes it so that through code, you can listen for when a Say Dialog starts and stops executing a Say Command. It was originally released here (please check the User Guide): https://drive.google.com/drive/folders/1FTRuqGiqi8xx_qF02O8tCD50S0ThABWn?usp=share_link

This PR has the code copied over from that, but with the appropriate changes applied. Stuff like making sure it's under Fungus's namespace, the SayDialog component is edited directly instead of subclassed, etc.

There's nothing that lets users respond to when Say Dialogs start or end without having to change their copy of the Fungus source directly.

### What is the new behavior?
Through a new class, SayDialogSignals, user code can listen for... well, you know at this point.

### Important Notes
- My change modifies the runtime execution/behaviour of Say Dialogs
- My change adds a demo to FungusExamples. The subfolder that has it is ListeningForSayDialog


breadnone:
This is super nice, so we can hook up events for 3rd parties to trigger animation for non-Fungus character like in 3D visualNovel or even Live2D character controller

